PR #56, opened 2016-07-28
Visibility weights

isullivan:
This branch changes most variable names that contain "flag" to "weights" or "vis_weights" as appropriate, making FHD terminology consistent with the uvfits definition and setting it up to weights column as actual weights in the future. This branch has been run through to the power spectrum, and the PS jackknife test shows the results are identical to the last standard reference run, v3.5-162-g19d3a2e_master.


nicholebarry:
I couldn't find anywhere you missed a name. I'll keep looking, though

PR #58, opened 2016-08-23
Calibration sim

nicholebarry:
Bare-bones optional calibration simulation setup. Includes two new files...

calibration_simulation_setup
- reverts weights to 1
- reads-in input model specified in keyword, or generates input model on the fly. 
- saves input model to sim_outputs directory
- reads-in optional eor simulation file and adds to visibilities
- reads-in optional visibility noise, or generates them on the fly.

vis_noise_simulation
- generates thermal noise on the visibilities
- saves noise to new sim_outputs directory

Since the cluster is down, I cannot test this. There are a few modifications to flagging that I don't believe are necessary, but I can't be sure till the cluster is back up. I did not include anything special in calibration (i.e. did not include "perfect" calibration option), since calibrations can be transferred anyways.

Would definitely like feedback on error handling.

Note: history includes versions dumps, so beware


nicholebarry:
Main options have been tested and are working.

- line 171 of vis_calibrate was removed. If no flagging is done, this line breaks.

nicholebarry:
Ian, can I get an approval?

fhd_core/calibration/vis_calibrate.pro
```diff
@@ -168,7 +168,7 @@ FOR iter=0,calibration_flag_iterate DO BEGIN
     t2+=t3_a-t2_a
     
     IF Keyword_Set(flag_calibration) THEN vis_calibration_flag,obs,cal,n_tile_cut=n_tile_cut,_Extra=extra
-    IF n_tile_cut EQ 0 THEN BREAK
+
```
isullivan: Why was this line removed?

nicholebarry:
Actually, I'm going to rename calibration_sim functions to in-situ_sim. Calibration is a bit of a misnomer as this functionality can be used for non-cal sims.

Any objections?

fhd_core/calibration/vis_calibrate.pro
isullivan: Right, but it should just break out of the FOR loop (not crash), which is exactly what you want in that case. This loop should only run a single time in any case, unless calibration_flag_iterate has been set. It looks like what it really needs is n_tile_cut=0 set just before the loop, so that it is defined. Or, change it to IF Keyword_Set(n_tile_cut) THEN BREAK

nicholebarry: Ah! You are right, obvious in retrospect. Since in-situ sim skips a flagging procedure (vis_flag_basic), n_tile_cut isn't defined and breaks on this line. Will change to keyword_set.

Issue #68, opened 2017-03-02
Extended sources not being read into FHD


rlbyrne:
After creating a source catalog from an FHD deconvolution run, running firstpass with that catalog does not treat the extended sources as such. 
![1130789944_uniform_sources_i](https://cloud.githubusercontent.com/assets/12840057/23488933/73d4ae32-fea4-11e6-9a22-48a94ed99ac4.png)
[1130789944_settings.txt](https://github.com/EoRImaging/FHD/files/825603/1130789944_settings.txt)

[firstpass.o73361.1.txt](https://github.com/EoRImaging/FHD/files/825608/firstpass.o73361.1.txt)

rlbyrne:
The uvfits and metafits files for this obsid are located on MIT in the directory /nfs/eor-05/r1/EoRuvfits/jd2457332v5_1/

isullivan:
I've been working on this today, and have an update if not a solution. The full extended source list is making it in to calibration and the visibilities used for calibration, but the extended source information is being lost from the source list after that. I haven't tracked down exactly where it is being lost, but at least if you set `return_cal_visibilities=1` and have no extra source modeling, the gridded model visibilities (and residual) should be correct. I still need to investigate the other cases, and track down where the extended pointer is being lost.

isullivan:
I believe I found the bug and fixed it. The changed code is part of pull request #79

PR #70, opened 2017-03-30
Beam clipping options

isullivan:
We need to finish testing before merging in to master! See Issue #68.


nicholebarry:
Update the dictionary.md as well. Perhaps we can add a "debug" section with a warning that those keywords can come and go, but it definitely will help me (at least) during heavier testing.

PR #75, opened 2017-03-30
Actually enable flagging individual time steps in firstpass

isullivan:


nicholebarry:
Needs definition for n_baselines. Otherwise, it works. I hardcoded removing the time_use overwrite at the end of transferring the weights to check.

PR #82, opened 2017-04-24
Healpix coordinate transformation for super-horizon pixels

isullivan:


aelanman:
This resolved the projection issues I was having when using a diffuse model stored as a full sphere healpix map, and didn't appear to break anything else.

isullivan:
I'm glad this fixed your bug. I'm seeing a slight difference in the final power spectrum, so I'm going to do a few more tests before merging this back into master.

PR #88, opened 2017-06-13
Enables the GSM to be used in simulations

aelanman:
Changes are also compatible with other diffuse models saved as healpix maps.


simulation/simulation_wrappers/sim_versions_wrapper.pro
```diff
@@ -394,55 +534,56 @@ endif
...
+;    allow_sidelobe_model_sources=allow_sidelobe_model_sources,$
+;    allow_sidelobe_cal_sources=allow_sidelobe_cal_sources,$
+;    psf_resolution=psf_resolution,$
+;    kbinsize=kbinsize,$
+;    restore_last=restore_last,$
+;    max_sources=max_sources,$
+;    max_model_sources=max_model_sources,$
+;    max_calibration_sources=max_calibration_sources,$
+;    dft_threshold=dft_threshold,$
+;    diffuse_model=diffuse_model,$
```
isullivan: Just delete these lines, don't comment them out. We can always recover them with git if you want them back later.

aelanman:
Does it look any better now?

isullivan:
Thanks for cleaning up the history.

PR #90, opened 2017-07-27
/readonly for uvfits read

nicholebarry:
To get around some permissions issues, I set a readonly flag to the header/uvfits reader. I don't think we'd ever be in a situation where we would like anything other than readonly, but I wanted to check before I pushed to master.


isullivan:
That should be fine. We don't ever want to be in a situation where we are writing back to the original uvfits files!

PR #97, opened 2017-10-25
EoR bubble simulations

aelanman:
This branch began as just adding EoR bubble simulation capabilities to FHD, but I've also used it to restructure my wrappers a little and make a couple changes to how healpix indices are selected (so it doesn't default to the EoR1/EoR0 fields).

I've stopped tracking the wrappers submit_sims.sh, sim_versions_wrapper.pro, and test_idl.sh.


fhd_core/HEALPix/healpix_cnv_generate.pro
```diff
@@ -36,7 +36,9 @@ ENDIF ELSE radius=hpx_radius
 
 ;check if a string, if it is assume it is a filepath to a save file with the desired indices 
 ; (will NOT be over-written with the indices)
-IF Keyword_Set(restrict_hpx_inds) AND (size(restrict_hpx_inds,/type) NE 7) THEN restrict_hpx_inds=observation_healpix_inds_select(obs)
+IF Keyword_Set(restrict_hpx_inds) AND (size(restrict_hpx_inds,/type) NE 7) THEN BEGIN
+      IF restrict_hpx_inds NE 0 THEN  restrict_hpx_inds=observation_healpix_inds_select(obs)
+ENDIF
```
isullivan: This change should be removed. `Keyword_Set(restrict_hpx_inds)` already returns `False` if `restrict_hpx_inds=0`

fhd_core/HEALPix/healpix_snapshot_cube_generate.pro
```diff
@@ -61,9 +61,10 @@ PRO healpix_snapshot_cube_generate,obs_in,status_str,psf_in,cal,params,vis_arr,v
   ENDIF ELSE  psf_out=beam_setup(obs_out,0,antenna_out,/no_save,psf_resolution=ps_psf_resolution,/silent,_Extra=extra)
   
   beam_arr=beam_image_cube(obs_out,psf_out,n_freq=n_freq_use,beam_mask=beam_mask,/square,beam_threshold=beam_threshold)
-  
+  if N_Elements(hpx_radius) EQ 0 then hpx_radius=FoV_use/sqrt(2.)
+  if N_Elements(restrict_hpx_inds) EQ 0 then restrict_hpx_inds=0
```
isullivan: Is it necessary to set `restrict_hpx_inds` here?

fhd_core/beam_modeling/beam_setup.pro
```diff
@@ -24,14 +24,14 @@ IF Keyword_set(transfer_psf) then begin
...
-      '/metadata/' + obs.obsname + '_obs.sav not found during psf transfer. Beam_integral needed in PS.'
+      '/metadata/' + obs.obsname + '_obs.sav not found during psf transfer. primary beam_sq_area needed in PS.'
       
     psf = getvar_savefile(transfer_psf + '/' + obs.obsname + '_beams.sav','psf')
     if file_test(transfer_psf + '/' + obs.obsname + '_obs.sav') then $
       obs_restore = getvar_savefile(transfer_psf + '/' + obs.obsname + '_obs.sav','obs')
     if file_test(file_dirname(transfer_psf) + '/metadata/' + obs.obsname + '_obs.sav') then $
       obs_restore = getvar_savefile(file_dirname(transfer_psf) + '/metadata/' + obs.obsname + '_obs.sav','obs')
-    obs.beam_integral = pointer_copy(obs_restore.beam_integral)
+    obs.primary_beam_sq_area = pointer_copy(obs_restore.primary_beam_sq_area)
```
isullivan: Please check with Bryna that it is OK to change this name, for Eppsilon.

fhd_core/beam_modeling/beam_setup.pro
```diff
@@ -156,7 +159,8 @@ FOR pol_i=0,n_pol-1 DO BEGIN
             t_beam_power+=Systime(1)-t_bpwr
             t_bint=Systime(1)
             ;divide by psf_resolution^2 since the FFT is done at a different resolution and requires a different normalization
-            beam_int+=baseline_group_n*Total(Abs(psf_base_superres)^2,/double)/psf_resolution^2. 
+            beam_int+=baseline_group_n*Total(psf_base_superres,/double)/psf_resolution^2.
```
isullivan: Should the beam integral be over the absolute value of the beam?

fhd_core/gridding/visibility_degrid.pro
```diff
@@ -59,11 +59,11 @@ IF n_conj GT 0 THEN BEGIN
...
 
-xcen=frequency_array#kx_arr
-ycen=frequency_array#ky_arr
+xcen=Float(frequency_array#kx_arr)
+ycen=Float(frequency_array#ky_arr)
  
-x_offset=Floor((xcen-Floor(xcen))*psf_resolution) mod psf_resolution    
-y_offset=Floor((ycen-Floor(ycen))*psf_resolution) mod psf_resolution 
+x_offset=Fix(Floor((xcen-Floor(xcen))*psf_resolution) mod psf_resolution, type=12) ; type=12 is unsigned int
+y_offset=Fix(Floor((ycen-Floor(ycen))*psf_resolution) mod psf_resolution, type=12) ; type=12 is unsigned int
```
isullivan: Aren't all of these changes already in `master`?

fhd_core/gridding/visibility_grid.pro
```diff
@@ -312,8 +308,8 @@ FOR bi=0L,n_bin_use-1 DO BEGIN
         vis_n=n_xyf_bin
     ENDIF ELSE BEGIN
         rep_flag=0
-        IF model_flag THEN model_box=model_use[inds];*freq_norm[freq_i]
-        vis_box=vis_arr_use[inds];*freq_norm[freq_i]
+        IF model_flag THEN model_box=model_use[inds]
+        vis_box=vis_arr_use[inds]
```
isullivan: As above, aren't these changes in `master` already?

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -50,14 +50,8 @@ FOR ti=0,N_Elements(time_cut)<2-1 DO BEGIN
...
 
-IF Tag_exist(hdr,'freq_arr') THEN BEGIN
-    freq_res=hdr.freq_res
-    frequency_array=hdr.freq_arr
-ENDIF ELSE BEGIN
-    freq_res=hdr.freq_width
-    ;frequency_array=(findgen(n_freq)-(hdr.freq_ref_i-1))*freq_res+hdr.freq_ref ;FITS header indices start at 1
-    frequency_array=(findgen(n_freq)-(hdr.freq_ref_i))*freq_res+hdr.freq_ref ;LEAVE unchanged for now to allow comparison!
-ENDELSE
+freq_res=hdr.freq_res
```
isullivan: More frequency fix duplicates.

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -117,8 +111,8 @@ ENDIF
 freq_use=Lonarr(n_freq)+1
 tile_use=Lonarr(n_tile)+1
 
-kx_arr=params.uu#frequency_array
-ky_arr=params.vv#frequency_array
+kx_arr=Float(params.uu#frequency_array)
+ky_arr=Float(params.vv#frequency_array)
```
isullivan: More frequency fix duplicates.

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -185,15 +179,15 @@ ENDIF
 pol_names=['XX','YY','XY','YX','I','Q','U','V']
 healpix={nside:Long(nside),ind_list:String(ind_list),n_pix:Long(n_hpx),n_zero:Long(n_zero_hpx)}
 
-arr={tile_A:Long(tile_A),tile_B:Long(tile_B),bin_offset:Long(bin_offset),Jdate:meta.Jdate,freq:Float(frequency_array),fbin_i:Long(freq_bin_i),$
+arr={tile_A:Long(tile_A),tile_B:Long(tile_B),bin_offset:Long(bin_offset),Jdate:meta.Jdate,freq:Double(frequency_array),fbin_i:Long(freq_bin_i),$
```
isullivan: More frequency fix duplicates.

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -185,15 +179,15 @@ ENDIF
...
 struct={code_version:String(code_version),instrument:String(instrument),obsname:String(obsname),$
     dimension:Float(dimension),elements:Float(elements),nbaselines:Long(nbaselines),dft_threshold:Float(dft_threshold),double_precision:double_precision,$
     kpix:Float(kbinsize),degpix:Float(degpix),obsaz:meta.obsaz,obsalt:meta.obsalt,obsra:meta.obsra,obsdec:meta.obsdec,$
     zenra:meta.zenra,zendec:meta.zendec,obsx:meta.obsx,obsy:meta.obsy,zenx:meta.zenx,zeny:meta.zeny,$
     phasera:meta.phasera,phasedec:meta.phasedec,orig_phasera:meta.orig_phasera,orig_phasedec:meta.orig_phasedec,$
     n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:n_time_cut,$
-    n_vis:Long(n_vis),n_vis_in:Long(n_vis_in),n_vis_raw:Long(n_vis_raw),nf_vis:Long(n_vis_arr),beam_integral:Ptrarr(4),pol_names:pol_names,$
+    n_vis:Long(n_vis),n_vis_in:Long(n_vis_in),n_vis_raw:Long(n_vis_raw),nf_vis:Long(n_vis_arr),primary_beam_area:Ptrarr(4),primary_beam_sq_area:Ptrarr(4),pol_names:pol_names,$
```
isullivan: As above, check with Bryna about the name change, and coordinate with any needed changes to Eppsilon

fhd_core/source_modeling/fhd_diffuse_model.pro
```diff
@@ -52,6 +52,17 @@ IF size(diffuse_spectral_index,/type) EQ 10 THEN BEGIN ;check if pointer type
...
+;if keyword_set(select_radius) THEN BEGIN
+;    ang2vec,obs.obsdec,obs.obsra,cen_coords,/astro
+;    Query_disc,nside,cen_coords,select_radius,inds_select,npix_sel,/deg
+;    pix2vec_ring, nside, inds_select, pix_coords
+;    vec2ang,pix_coords,pix_dec,pix_ra,/astro
+;    IF coord_use EQ 'galactic'   THEN glactc,pix_ra,pix_dec,2000.,pix_ra,pix_dec,2, /degree
+;    IF coord_use EQ 'equatorial' THEN Hor2Eq,pix_dec,pix_ra,obs.JD0,pix_ra,pix_dec,lat=obs.lat,lon=obs.lon,alt=obs.alt,precess=1,/nutate
+;    ang2vec, pix_dec, pix_ra, vecs, /astro
+;    vec2pix_ring, nside, vecs, hpx_inds
+;ENDIF
```
isullivan: Don't add commented-out code to `master`. If this is not added, then also don't add the keyword `select_radius` above.

fhd_core/visibility_manipulation/vis_header_extract.pro
```diff
@@ -16,12 +16,11 @@ n_pol=sxpar(header,'naxis3') ;4 columns are xx, yy, xy, yx
...
 freq_res=Double(sxpar(header,'cdelt4')) ;40000 
-freq_res=Double(Floor(freq_res/1000))*1000.   ; Round down to kHz
+freq_res=Double(freq_res)
 IF freq_res EQ 0 THEN print,'WARNING: Invalid frequency resolution in uvfits header! Uses header keyword cdelt4'
-freq_ref_i=Double(sxpar(header,'crpix4')-1)  ;368-1 (Remember, FITS indices start from 1, IDL indices start from 0)
+freq_ref_i=sxpar(header,'crpix4')-1  ;368-1 (Remember, FITS indices start from 1, IDL indices start from 0)
 date_obs=sxpar(header,'date-obs')
-frequency_array=(findgen(n_freq)-freq_ref_i)*freq_res+freq_ref
-frequency_array=Float(Floor(frequency_array/1000)*1000)   ; Round down to kHz 
+frequency_array=(dindgen(n_freq)-freq_ref_i)*freq_res+freq_ref
```
isullivan: More frequency fix duplicates.

simulation/array_simulator.pro
```diff
@@ -51,7 +51,27 @@ PRO array_simulator,vis_arr,vis_weights,obs,status_str,psf,params,jones,error=er
   IF ~silent THEN print,'Calculating beam model'
   psf=beam_setup(obs,status_str,file_path_fhd=file_path_fhd,restore_last=0,silent=silent,timing=t_beam,no_save=0,_Extra=extra)
   IF Keyword_Set(t_beam) THEN IF ~silent THEN print,'Beam modeling time: ',t_beam
-  
+ 
+  ;; Get a selection radius based on the primary beam width  TODO --- Only do this if include_catalog_sources == 0, in which case get it from generate_source_cal_list
+  if not keyword_set(beam_threshold) then beam_threshold = 0.05
+  if keyword_set(allow_sidelobe_sources) THEN beam_threshold = 0.01
```
isullivan: This is not a good place to set these defaults. Based on the code below, these lines should simply be removed from this file.

simulation/array_simulator.pro
```diff
@@ -51,7 +51,27 @@ PRO array_simulator,vis_arr,vis_weights,obs,status_str,psf,params,jones,error=er
...
+  IF N_Elements(beam_arr) LT (n_pol<2) THEN BEGIN 
+      beam_arr=Ptrarr(n_pol<2)
+      FOR pol_i=0,(n_pol<2)-1 DO beam_arr[pol_i]=Ptr_new(beam_image(psf,obs,pol_i=pol_i,square=0)>0.)
+  ENDIF
+  beam=fltarr(dimension,elements)
+  FOR pol_i=0,(n_pol<2)-1 DO beam+=*beam_arr[pol_i]^2.
+  beam=Sqrt(beam/(n_pol<2))
+  beam_primary_i=region_grow(beam,dimension/2.+dimension*elements/2.,threshold=[Max(beam)/2.<beam_threshold,Max(beam)>1.])
+  boundary=find_boundary(beam_primary_i,xsize=dimension, ysize=elements,perim_area=perim_area)
+  select_radius = (degpix*sqrt(perim_area)/2 + 3.)   ;Primary beam radius + 3 degrees
```
isullivan: This new code needs to all be in a separate function. Then you can safely set the defaults for beam_threshold you use above, though you should think about what you are doing if `allow_sidelobe_sources` is set. This calculation will expand the beam, but the radius it produces will *not* actually include the sidelobes. So, either treat the sidelobes properly or (most likely) simply ignore the keyword `allow_sidelobe_sources` for this calculation.

simulation/array_simulator.pro
```diff
@@ -162,7 +182,7 @@ PRO array_simulator,vis_arr,vis_weights,obs,status_str,psf,params,jones,error=er
   IF Keyword_Set(snapshot_healpix_export) THEN begin
     IF ~Keyword_Set(n_avg) THEN n_avg=1
     IF Keyword_Set(snapshot_recalculate) THEN status_str.healpix_cube=(status_str.hpx_even=(status_str.hpx_odd=0))
-    healpix_snapshot_cube_generate,obs,status_str,psf,cal,params,vis_arr,/restrict_hpx_inds,$
+    healpix_snapshot_cube_generate,obs,status_str,psf,cal,params,vis_arr,hpx_radius=Float(select_radius),$
```
isullivan: Removing the `/restrict_hpx_inds` keyword will change the behavior of `array_simulator` for all other users. We perhaps don't want that hardcoded, but it should still be passed in and kept as the default.

isullivan: `select_radius` should be returned as a `Float` from the (new) function that calculates it. The extra conversion to `Float` is unnecessary and confusing.

simulation/vis_simulate.pro
```diff
@@ -19,15 +20,16 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
...
+ 
   ;Construct model visibilities. Start by building a model u-v-f cube
   if keyword_set(include_catalog_sources) then begin
     catalog_source_array=generate_source_cal_list(obs,psf,catalog_path=catalog_file_path,_Extra=extra)
     if n_elements(source_array) gt 0 then source_array = [source_array, catalog_source_array] else source_array = catalog_source_array
   endif    
   n_sources=N_Elements(source_array)
   print, 'n_sources: '+string(n_sources)
-  skymodel=fhd_struct_init_skymodel(obs,source_list=source_array,catalog_path=catalog_file_path,diffuse_model=diffuse_model, return_cal=0,_Extra=extra)
+  skymodel=fhd_struct_init_skymodel(obs,source_list=source_array,catalog_path=catalog_file_path,diffuse_model=diffuse_model, beam_threshold=beam_threshold, return_cal=0,_Extra=extra)
```
isullivan: I don't think `beam_threshold` is defined anywhere else, so it shouldn't be passed in.

simulation/vis_simulate.pro
```diff
@@ -157,11 +159,21 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
...
+         endif else model_uvf_arr = Pointer_copy(bubble_uvf)
+       ENDIF
+    ENDIF 
+
     ;; If diffuse model is included, calculate its contribution and add to the source_model_uv_arr
     IF Keyword_Set(diffuse_model) THEN BEGIN
       IF file_test(diffuse_model) EQ 0 THEN diffuse_model=(file_search(diffuse_model+'*'))[0]
          print,"Reading diffuse model file: "+diffuse_model
-      diffuse_model_uv=fhd_diffuse_model(obs,jones,skymodel,spectral_model_arr=diffuse_spectral_model_uv,/uv_return,/diffuse_units_kelvin,model_filepath=diffuse_model,_Extra=extra)
+      diffuse_model_uv=fhd_diffuse_model(obs,jones,skymodel,spectral_model_arr=diffuse_spectral_model_uv, select_radius=select_radius,/uv_return,/diffuse_units_kelvin,model_filepath=diffuse_model,_Extra=extra)
```
isullivan: If you remove `select_radius` from `fhd_diffuse_model` above as I requested, you will need to remove it here, too.

simulation/vis_simulate.pro
```diff
@@ -176,7 +188,7 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
         ;; if there is also a uvf cube, add the uv from the sources to the cube at each freq.
 	IF size(*source_model_uv_arr[0],/type) eq 6 then begin     ;; If array is complex
 	        FOR pol_i=0,n_pol-1 DO BEGIN
-			*model_uvf_arr[pol_i]+=rebin_complex(*source_model_uv_arr[pol_i],dimension,elements,n_freq,/sample)
+			*model_uvf_arr[pol_i]+=rebin_complex(*source_model_uv_arr[pol_i],dimension,elements,n_freq)
```
isullivan: The `/sample` keyword should remain. That is a more correct calculation, unless you have a very good reason for removing it.

simulation/vis_simulate.pro
```diff
@@ -238,15 +250,17 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
...
-         ;       polarization=pol_i,_Extra=extra)
-        ;ENDFOR
+        ;  this_model_ptr=vis_source_model(skymodel,obs,status_str,psf,params,this_vis_weight_ptr,model_uv_arr=this_model_uv,$
+        ;    timing=model_timing,silent=silent,error=error,_Extra=extra)
+          model_timing=0.0
+          FOR pol_i=0,n_pol-1 DO BEGIN
+              (*vis_model_arr[pol_i])[fi,*] = (*(visibility_degrid(*this_model_uv[pol_i],this_vis_weight_ptr[pol_i],obs,psf,params,timing=timing,silent=silent,$
+                  polarization=pol_i,_Extra=extra)))[fi,*]
+               model_timing += timing
+          ENDFOR    ; vis_source_model uses DFT, which works for point source models. Since the uvf cube was already FFTed, just run visibility_degrid.
```
isullivan: Please move this comment up to the start of the new block of code. I don't like leaving commented-out code in, but that was there before and is not your responsibility to fix right now.

simulation/eor_bubble_sim.pro
```diff
@@ -0,0 +1,90 @@
+FUNCTION eor_bubble_sim, obs, jones, select_radius=select_radius, bubble_fname=bubble_fname, beam_threshold=beam_threshold, allow_sidelobe_sources=allow_sidelobe_sources, dat=dat 
+
+;Opening an HDF5 file and extract relevant data
+if keyword_set(bubble_fname) THEN hdf5_fname = bubble_fname ELSE hdf5_fname = '/users/alanman/data/alanman/BubbleCube/TiledHpxCubes/mwa_gaussian.hdf5'
+if not keyword_set(beam_threshold) then beam_threshold = 0.05
+if keyword_set(allow_sidelobe_sources) THEN beam_threshold = 0.01
+if not keyword_set(select_radius) THEN  select_radius = 20     ; Degrees
+
+dimension=obs.dimension
+elements= obs.dimension
```
isullivan: `elements = obs.elements`?

simulation/eor_bubble_sim.pro
```diff
@@ -0,0 +1,90 @@
...
+if n_elements(dat) eq 0 then dat = H5D_READ(dset_id_eor, FILE_SPACE=dpsace_id_eor)  ; temporary --- allow for the data to be reused if already loaded
+print, 'HDF5 reading time = ', systime(/seconds) - t0, ' seconds'
+
+; Interpolate in frequency:
+dat_interp = Fltarr(obs.n_freq,npix_sel)
+t0=systime(/seconds)
+for hpx_i=0,npix_sel-1 DO dat_interp[*,hpx_i] = Interpol(dat[freq_inds,inds_select[hpx_i]],freq_hpx,freq_arr, /spline)
+print, 'Frequency interpolation complete: ', systime(/seconds) - t0
+hpx_arr = Ptrarr(obs.n_freq)
+for fi=0, obs.n_freq-1 DO hpx_arr[fi] = ptr_new(reform(dat_interp[fi,*]))
```
isullivan: Do you need to worry about the case when the observation has more frequencies than the healpix cube?

simulation/eor_bubble_sim.pro
```diff
@@ -0,0 +1,90 @@
...
+
+
+H5S_CLOSE, dspace_id_eor
+H5D_CLOSE, dset_id_eor
+H5F_CLOSE, f_id
+
+n_pol=4
```
isullivan: Should this be hard-coded, or passed in (with a default)?

fhd_core/gridding/visibility_degrid.pro
aelanman: I think I merged the doubles_frequencies (or whatever it was called) into eor_bubble prior to merging into master, so this is a duplicate of that.

isullivan: Yes, if you *merged* them in, then the changes will be duplicated. That will lead to some problems when you go to merge this into `master`. Since those changes are now in `master`, you could rebase on this branch and remove those commits, then rebase off of `master`.

fhd_core/beam_modeling/beam_setup.pro
aelanman: I've submitted an issue on eppsilon. We talked about this change back in August.

fhd_core/beam_modeling/beam_setup.pro
aelanman: In this case, beam_int leads to the primary_beam_area, and beam2_int gives primary_beam_sq_area. The latter is what I'm proposing we rename beam_integral in obs.

primary_beam_area as defined here gives the correct quantity for power spectrum normalization with the PAPER pipeline, so I need to have it in the obs.

simulation/array_simulator.pro
aelanman: Most of this is just a duplicate of what's done in generate_source_cal_list for source selection (which sets defaults for beam_threshold according to allow_sidelobe_sources). The only change is the estimate of the beam radius from that. Where should I put this to act as a separate function?

simulation/eor_bubble_sim.pro
aelanman: That hasn't come up so far... Will Interpol fail in that case?

simulation/eor_bubble_sim.pro
aelanman: I realize that just above it I have n_pol = obs.n_pol. I'm confused by this, because obs.n_pol seems to be 2, not 4. It should definitely be 4, since this sets the size of model_tmp which holds Stokes parameters.

simulation/vis_simulate.pro
aelanman: Unless I'm looking at the wrong file, the rebin_complex function defined in fhd_utils doesn't have a sample keyword.

simulation/vis_simulate.pro
aelanman: Move the comment "vis_source_model uses DFT..." up to line 256?

simulation/eor_bubble_sim.pro
isullivan: If obs.n_pol is 2, that means there are only 2 polarizations of data present. You should definitely use obs.n_pol to be consistent with the rest of the data pipeline, but you could ensure that your simulation always has 4 polarizations if you need that.

simulation/array_simulator.pro
isullivan: Just write a new function, called something like `select_beam_pixels` and put all of this code in it. The new function should go in `fhd_utils`. Please also see my recent commit that changed `generate_source_cal_list`. You will want to apply those changes to this code as well.

fhd_core/beam_modeling/beam_setup.pro
isullivan: Yes, I recall talking about it, and I have no problem with the change. As long as it's compatible with Eppsilon, it's fine.

simulation/eor_bubble_sim.pro
isullivan: I think your array indexing will just crash in that case. It is probably some of work to write code that supports this case, so you might just check for that and issue a warning to catch it earlier, instead of letting people get an error message and thinking there's a bug.

simulation/array_simulator.pro
aelanman: Do you want the new function to do just the primary_beam width calculation, or do everything that's done in generate_source_cal_list and keep the width estimation in array_simulator?

simulation/eor_bubble_sim.pro
aelanman: I've tested this by reducing the size of freq_hpx and the frequency axis of dat while leaving the observation frequencies, and it didn't have any issues.
The interpolation step should ensure that the cube's frequency axis matches has length obs.n_freq, regardless of how many are in the healpix cube to begin with.

bhazelton:
@aelanman I'd like to see you make the appropriate eppsilon changes and submit a PR for eppsilon as well. Then we can run tests with both repos to make sure the change works across the handoff and we can coordinate merging the PRs.

bhazelton:
Ok, the necessary eppsilon changes have been merged in so that this should just work and it is still backwards compatible for old cubes. So that's no longer a block as far as I'm concerned.

It'd be good to run a ps difference test with this branch and eppsilon to make sure it all works, but I don't expect any problems from the eppsilon side.

aelanman:
@isullivan  Are there any further changes you want made, or do you think this is ready for merging?

simulation/eor_bubble_sim.pro
```diff
@@ -0,0 +1,84 @@
+FUNCTION eor_bubble_sim, obs, jones, select_radius=select_radius, bubble_fname=bubble_fname, beam_threshold=beam_threshold, allow_sidelobe_sources=allow_sidelobe_sources, dat=dat 
+
+;Opening an HDF5 file and extract relevant data
+if keyword_set(bubble_fname) THEN hdf5_fname = bubble_fname ELSE hdf5_fname = '/users/alanman/data/alanman/BubbleCube/TiledHpxCubes/mwa_gaussian.hdf5'
```
nicholebarry: Please remove hardcoding of your path, and add an error message if the bubble_fname fails a file_test.

simulation/eor_bubble_sim.pro
```diff
@@ -0,0 +1,84 @@
...
+freq_inds = where((freq_hpx GT lim[0]) and (freq_hpx LT lim[1]) )
+freq_hpx = freq_hpx[freq_inds]
+nfreq_hpx = n_elements(freq_hpx)
+
+;; Extract only these healpix indices from the file.
+;H5S_SELECT_ELEMENTS, dspace_id_eor, hpx_inds, /RESET
+
+print, "Reading HDF5 file with EoR Healpix Cube"
+t0 = systime(/seconds)
+if n_elements(dat) eq 0 then dat = H5D_READ(dset_id_eor, FILE_SPACE=dpsace_id_eor)  ; temporary --- allow for the data to be reused if already loaded
```
nicholebarry: Comment says this is temporary?

simulation/eor_bubble_sim.pro
```diff
@@ -0,0 +1,84 @@
...
+model_stokes_arr = healpix_interpolate(hpx_arr,obs,nside=nside,hpx_inds=inds_select,/from_kelvin)
+print, 'Hpx_interpolate timing: ', systime(/seconds) - t0
+
+FOR fi=0, obs.n_freq-1 do begin    ; 30 seconds for 203 channels
+   model_tmp=*model_stokes_arr[fi]
+   model_tmp=Ptrarr(n_pol,/allocate)
+   *model_tmp[0] = *model_stokes_arr[fi]
+   FOR pol_i=1,n_pol-1 DO *model_tmp[pol_i]=Fltarr(obs.dimension,obs.elements)
+   model_arr = stokes_cnv(model_tmp, jones, /inverse)
+   Ptr_free, model_tmp
```
nicholebarry: It seems as though model_tmp is being redefined multiple times. Can you make this more succinct?

simulation/vis_simulate.pro
```diff
@@ -49,9 +51,9 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
       ENDELSE
 
     endfor
-    IF ~Keyword_Set(no_save) THEN save, file=init_beam_filepath, beam2_xx_image, beam2_yy_image, obs
-    undefine_fhd, beam2_xx_image, beam2_yy_image,beam_arr
-    
+;    IF ~Keyword_Set(no_save) THEN save, file=init_beam_filepath, beam2_xx_image, beam2_yy_image, obs
```
nicholebarry: I can't see a reason to remove this line

simulation/vis_simulate.pro
```diff
@@ -157,6 +159,16 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
       
     endif ;; end if n_elements(model_image_cube) gt 0 or n_elements(model_uvf_cube) gt 0 or keyword_set(include_eor)
     
+    ;; If bubble file is included, add to the model_uvf_arr
+    IF Keyword_Set(bubble_fname) THEN BEGIN 
+       bubble_uvf = eor_bubble_sim(obs, jones, select_radius=select_radius, bubble_fname=bubble_fname)
+       IF Min(Ptr_valid(bubble_uvf)) GT 0 THEN BEGIN
+         IF n_elements(model_uvf_arr) GT 0 then begin
+           FOR pol_i=0,n_pol-1 DO *model_uvf_arr[pol_i] += *bubble_uvf[pol_i];*uv_mask_use
+         endif else model_uvf_arr = Pointer_copy(bubble_uvf)
```
nicholebarry: Please make sure that the bubble_uvf is in the right units. I've had trouble in the past (and still do!) about whether or not we are using XX+YY/2 or XX+YY. I don't know what the right answer is, but we should definitely look into it. We should compare directly to another simulation and see if we're off by a factor of 2...

simulation/vis_simulate.pro
```diff
@@ -238,15 +250,18 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
...
-         ;   vis_model_arr[pol_i]=visibility_degrid(*this_model_uv[pol_i],this_vis_weight_ptr[pol_i],obs,psf,params,silent=silent,$
-         ;       polarization=pol_i,_Extra=extra)
-        ;ENDFOR
+        ;  this_model_ptr=vis_source_model(skymodel,obs,status_str,psf,params,this_vis_weight_ptr,model_uv_arr=this_model_uv,$
+        ;    timing=model_timing,silent=silent,error=error,_Extra=extra)
+          model_timing=0.0
+    ; vis_source_model uses DFT, which works for point source models. Since the uvf cube was already FFTed, just run visibility_degrid.
+          FOR pol_i=0,n_pol-1 DO BEGIN
+              (*vis_model_arr[pol_i])[fi,*] = (*(visibility_degrid(*this_model_uv[pol_i],this_vis_weight_ptr[pol_i],obs,psf,params,timing=timing,silent=silent,$
+                  polarization=pol_i,_Extra=extra)))[fi,*]
```
nicholebarry: I don't think this is applicable for broader situations. `vis_source_model` will happily dft/add point sources, diffuse model, or a galaxy model to your input bubble uvf, but will skip directly to `visibility_degrid` if none of those things are specified. I definitely can see people adding things other than eor in the future, so let's roll this back.

simulation/eor_bubble_sim.pro
aelanman: It doesn't need to be temporary... I was doing some tests and made it so eor_bubble_sim could accept the data object instead of re-reading it to save on time. It's not currently used, but might be useful. I can remove the option if you'd prefer.

simulation/vis_simulate.pro
aelanman: The saved beam models were hogging up a lot of storage for me, so I cut the line since I didn't need them. :)
I'll put it back.

simulation/eor_bubble_sim.pro
aelanman: I think I got this... there was an extra definition in there.

simulation/vis_simulate.pro
aelanman: By my reading of stokes_cnv.pro, the convention in FHD seems to be the following:
``` I = XX + YY
XX = (I + Q)/2
YY = (I - Q)/2 
```
The bubble_uvf array obtained from eor_bubble_sim is in the (xx,yy,xy,yx) basis, having been converted with stokes_cnv(.... ,/inverse). It should therefore obey the same Stokes convention as the rest of FHD.

PR #98, opened 2017-10-25
Convert frequency_array to double precision

isullivan:
Note that `frequency_array` - derived large arrays are converted back to `float` precision after their derivation.


aelanman:
I've tested this with a small point+EoR simulation with uncompressed (97.5 kHz) channels and got visibility data with evenly spaced frequencies. This seems to have resolved the truncation issue satisfactorily.

PR #114, opened 2018-02-13
Eor bubble fixes

aelanman:
The most significant change in this branch is that I _correctly_ form the uvf cube in eor_bubble_sim so that the degridding step actually loops over frequency. For the sake of computational speed, I also switch back to using visibility_degrid instead of its wrapper (vis_source_model) in vis_simulate. There was some discussion back in January about the differences of these two, and it might be worth reexamining vis_simulate.

### DIFF
--- a/simulation/eor_bubble_sim.pro
+++ b/simulation/eor_bubble_sim.pro
@@ -32,14 +32,20 @@ freq_inds = where((freq_hpx GT lim[0]) and (freq_hpx LT lim[1]) )
 freq_hpx = freq_hpx[freq_inds]
 nfreq_hpx = n_elements(freq_hpx)
 
+; making the coord array for h5s_select
+;coord_arr = array_indices(make_array(nfreq_hpx,npix_sel),findgen(nfreq_hpx*npix_sel))
+;coord_arr[1,*] = inds_select[coord_arr[1,*]]
+
 ;; Extract only these healpix indices from the file.
-;H5S_SELECT_ELEMENTS, dspace_id_eor, hpx_inds, /RESET
+; Somehow this ended up being slower than reading in the full array, so leave it out for now.
+;H5S_SELECT_ELEMENTS, dspace_id_eor, coord_arr, /RESET
 
 print, "Reading HDF5 file with EoR Healpix Cube"
 t0 = systime(/seconds)
 if n_elements(dat) eq 0 then dat = H5D_READ(dset_id_eor, FILE_SPACE=dpsace_id_eor)
 print, 'HDF5 reading time = ', systime(/seconds) - t0, ' seconds'
 
+; dat now contains zeros for the nonselected healpix indices, but retains the full shell shape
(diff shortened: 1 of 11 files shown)


simulation/eor_bubble_sim.pro
```diff
@@ -55,27 +55,24 @@ H5S_CLOSE, dspace_id_eor
...
-print, 'Hpx_interpolate timing: ', systime(/seconds) - t0
+print, 'Healpix_interpolate timing: ', systime(/seconds) - t0
 
-FOR fi=0, obs.n_freq-1 do begin    ; 30 seconds for 203 channels
+FOR pol_i=0, n_pol-1 DO *model_uv_arr[pol_i]=Fltarr(obs.dimension,obs.elements,obs.n_freq)
+FOR fi=0, obs.n_freq-1 do begin
    model_tmp=Ptrarr(n_pol,/allocate)
    *model_tmp[0] = *model_stokes_arr[fi]
-   FOR pol_i=1,n_pol-1 DO *model_tmp[pol_i]=Fltarr(obs.dimension,obs.elements)
+   *model_tmp[1] = fltarr(obs.dimension, obs.elements)
```
isullivan: This will cause an error if `n_pol` is anything other than 2. I'd suggest reverting this one line

simulation/vis_simulate.pro
```diff
@@ -263,18 +263,18 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
...
+    ; vis_source_model is inexplicably taking a long time. Since the uvf cube was already FFTed, just run visibility_degrid.
+          FOR pol_i=0,n_pol-1 DO BEGIN
+              (*vis_model_arr[pol_i])[fi,*] = (*(visibility_degrid(*this_model_uv[pol_i],this_vis_weight_ptr[pol_i],obs,psf,params,timing=timing,silent=silent,$
+                  polarization=pol_i,_Extra=extra)))[fi,*]
+               model_timing += timing
+          ENDFOR
           print, 'model loop num, timing(s):'+ number_formatter(fi) + ' , ' + number_formatter(model_timing)
           
-          for pol_i=0,n_pol-1 do (*vis_model_arr[pol_i])[fi,*] = (*this_model_ptr[pol_i])[fi,*]
+;          for pol_i=0,n_pol-1 do (*vis_model_arr[pol_i])[fi,*] = (*this_model_ptr[pol_i])[fi,*]
```
isullivan: If you are leaving commented-out code to make it easier to switch back later, please make sure it is all in one block and clearly marked. Also, please indent your comments to the same level as the surrounding code (but don't indent the commented-out code).

aelanman:
I've addressed your comments @isullivan . I switched back to the pol_i loop in eor_bubble_sim so that all pols are initialized to fltarr(dimension, elements), in case n_pol != 2. I removed the commented code.

I would still like to find a time to update and improve vis_simulate.

PR #117, opened 2018-02-20
Only convert HEALPix indices that are above the horizon

isullivan:
When converting from image pixels to HEALPix, this first first does a rough calculation of altitude and azimuth, and cuts any pixels with altitude less than 0.


bhazelton:
Looks good to me. I might consider adding a print statement in the if statement to let the user know what happened...

PR #123, opened 2018-04-18
Revised healpix_interpolate

aelanman:
This replaces the old healpix_interpolate function with a new method. The new method bins the healpix values into an orthoslant grid with a resolution that matches that of the input map, then either truncates or zero-pads in the uv plane to obtain the true orthoslant grid of the instrument. Kelvin to Jy/pix conversion is saved for the last step, if needed.

This removes the observed angular deviation found in the old healpix_interpolate, and behaves as expected when given gaussian healpix shells. I've also done some quick checks that it doesn't alter the hera beam model, since this relies on healpix_interpolate as well.

~Also included is an older set of fixes to my wrappers and to eor_bubble_sim, so that the latter makes a uvf cube with the correct form so the degridding step will loop over frequency. The version currently on the master branch doesn't loop over frequency in degridding because the uvf cube is incorrectly shaped, so the results are nonsensical.~ The wrappers have been moved to `pipeline_scripts`.

Update on 3/6/2020:
I've rebased and pushed the latest of this branch. I haven't been keeping up as closely with the more recent changes to FHD, but I think the changes I've made here may still be useful.


nicholebarry:
I'll set up a signal loss run on zenith for Phase I with this branch

bhazelton:
@nicholebarry we'd like to push on this PR, there are some tests relying on it but we want to make sure it's well behaved and passes the signal loss test. Have you had a chance to run it or can you tell us how to set one up?

isullivan:
I haven't looked at the code in detail yet, but I have run both this branch and the `healpix_binterp` branch through my power spectrum jackknife test. Both give a PS that is precisely equal to the PS created from the `master` branch, at least for a single zenith observation.

That seems encouraging, particularly for this branch where I didn't expect a change, but I had thought the `healpix_binterp` branch would give a slightly different power spectrum. Am I missing something, or should these changes only have an effect when you are processing a long integration?

aelanman:
It's been a couple years since I pushed changes to this branch... would anyone be willing to take a look? I've rerun an EoR signal simulation recently with comparable results to what I've seen before with this branch, so I can say that it works as I expect it to.

@rlbyrne @bhazelton @isullivan

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+n_hpx=nside2npix(nside)
+IF N_Elements(hpx_inds) EQ 0 THEN hpx_inds=Lindgen(n_hpx)
+hpx_res = sqrt(4*!Pi/(n_hpx))*!RaDeg
+dimension_hpx = ceil((obs.dimension*obs.degpix/hpx_res)*(1/2.))*2    ; Ensure even.
+elements_hpx = ceil((obs.elements*obs.degpix/hpx_res)*(1/2.))*2
+; give new dimension/elements to update_obs
+obs_hpx = fhd_struct_update_obs(obs,dimension=dimension_hpx, elements=elements_hpx, degpix=hpx_res)
+
+;apply_astrometry,obs_hpx, x_arr=meshgrid(dimension_hpx,elements_hpx,1), y_arr=meshgrid(dimension_hpx, elements_hpx, 2), ra_arr=ra_arr, dec_arr=dec_arr, /xy2ad
+;radec_i=where(Finite(ra_arr))
```
adampbeardsley: remove commented code?

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+image_mask[Ceil(xv_hpx),Floor(yv_hpx)] = 1
+image_mask[Floor(xv_hpx),Ceil(yv_hpx)] = 1      ; Cover the full circular region
+
+
+x_frac=1.-(xv_hpx-Floor(xv_hpx))
+y_frac=1.-(yv_hpx-Floor(yv_hpx))
+IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,degpix=obs.degpix,freq=obs.freq_center) $
+    ELSE pixel_area_cnv=((obs.degpix*!DtoR)^2.)/(4.*!Pi/n_hpx) ; (steradian/new pixel)/(steradian/old pixel)
+
+;;orthoslant_scale = pixel_area(obs_hpx, /relative)    ; Pixel areas to projected pixel areas.
```
adampbeardsley: remove?

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+
+FOR map_i=0,n_map-1 DO BEGIN
+    model_img = fltarr(dimension_hpx,elements_hpx)
+    IF Ptr_flag THEN hpx_vals=(*healpix_map[map_i])[hpx_i_use] ELSE hpx_vals=healpix_map[hpx_i_use] 
+;    hpx_vals*=pixel_area_cnv ;convert to Jy/pixel for the new Healpix pixels
```
adampbeardsley: remove

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+
+    for ni=0,n_hpx_use-1 DO model_img[round(xv_hpx[ni]),round(yv_hpx[ni])] += hpx_vals[ni]
+;;    model_img *= orthoslant_scale
```
adampbeardsley: remove

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+    RETURN,map_interp
+ENDIF
+xv_hpx=xv_hpx[hpx_i_use]
+yv_hpx=yv_hpx[hpx_i_use]
+
+image_mask=Fltarr(dimension_hpx,elements_hpx)
+image_mask[Floor(xv_hpx),Floor(yv_hpx)] = 1
+image_mask[Ceil(xv_hpx),Ceil(yv_hpx)] = 1
+image_mask[Ceil(xv_hpx),Floor(yv_hpx)] = 1
+image_mask[Floor(xv_hpx),Ceil(yv_hpx)] = 1      ; Cover the full circular region
```
adampbeardsley: I don't really understand what this mask is doing. Looks like it's used in line 70, but anywhere that mask is zero, the model_img should be zero anyway (I think?)

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+;;    model_img *= orthoslant_scale
+    model_uv = fft_shift(FFT(model_img * image_mask))
+
+    ; Zero pad or truncate to the true orthoslant dimension
+    dim_out = obs.dimension
+    ele_out = obs.elements
+    IF dim_out LT dimension_hpx THEN BEGIN
+        ; Truncate
+        model_uv_full = model_uv[dimension_hpx/2. - dim_out/2.: dimension_hpx/2. + dim_out/2.-1,$
+                                   elements_hpx/2. - ele_out/2.: elements_hpx/2. + ele_out/2.-1 ]
```
adampbeardsley: You check whether to truncate in one dimension, then assume you do the same in the other dimension. I don't know for sure if this is a safe assumption... if you have oblong fields/beams, you may have different situations?

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,92 @@
...
+    ENDIF ELSE BEGIN
+        ; Zero-pad
+        model_uv_full = Complexarr(dim_out,ele_out)
+        model_uv_full[dim_out/2. - dimension_hpx/2.: dim_out/2. + dimension_hpx/2.-1,$
+            ele_out/2. - elements_hpx/2.: ele_out/2. + elements_hpx/2. -1] = model_uv
+    ENDELSE
+
+    scale=1.0
```
adampbeardsley: I don't think this scale is necessary. Is it a placeholder for something else?

fhd_core/fhd_main.pro
```diff
@@ -253,8 +253,11 @@ IF Keyword_Set(export_images) THEN BEGIN
 ENDIF
 
 ;optionally export frequency-splt Healpix cubes
+;;; Temporarily --- Include options from simulation to select a circular region of healpix indices
```
adampbeardsley: Is this temporary choice is relevant?

fhd_core/HEALPix/healpix_bin.pro
aelanman: This step performs a horizon cut on the data. The previous two lines select only those pixels which fall within the dimension_hpx / elements_hpx box.

fhd_core/HEALPix/healpix_bin.pro
aelanman: Hm... I guess if the pixel sizes in each direction differ this could happen? This wasn't written for the simpler case -- are the orthoslant pixels larger or smaller than the healpix pixels (in solid angle)?

I'm not sure what the approach should be for that case, but it shouldn't depend on the beam.

fhd_core/fhd_main.pro
aelanman: I think this was me trying to start including these options in the FHD code for data analysis. It has no effect on simulations, so I'll just remove the change.

fhd_core/HEALPix/healpix_bin.pro
aelanman: Done.

fhd_core/HEALPix/healpix_bin.pro
aelanman: Done.

fhd_core/HEALPix/healpix_bin.pro
aelanman: Done.

fhd_core/HEALPix/healpix_bin.pro
aelanman: Done.

fhd_core/HEALPix/healpix_bin.pro
aelanman: It's removed now. I don't remember what it was for exactly, but it's not needed now.

nicholebarry:
@aelanman Can you do a brief check on this to make sure everything is still relevant?

Also, a little summary for me would be nice.

aelanman:
@nicholebarry I've rebased this branch and added some more details to the dictionary. This branch enables simulations of diffuse models (from "full-frequency healpix maps") to `vis_simulate.pro`.

This was the first end to end test of simulating a flat-spectrum EoR model from a sky model to visibilities to power spectra. It does this by selecting pixels from the maps (by default, to within the primary beam radius) and binning them in the 2D orthoslant grid. This is in contrast to the interpolation methods used by `healpix_interp` for setting up beam models.

fhd_core/HEALPix/healpix_bin.pro
```diff
@@ -0,0 +1,84 @@
...
+image_mask=Fltarr(dimension_hpx,elements_hpx)
+image_mask[Floor(xv_hpx),Floor(yv_hpx)] = 1
+image_mask[Ceil(xv_hpx),Ceil(yv_hpx)] = 1
+image_mask[Ceil(xv_hpx),Floor(yv_hpx)] = 1
+image_mask[Floor(xv_hpx),Ceil(yv_hpx)] = 1      ; Cover the full circular region
+
+
+x_frac=1.-(xv_hpx-Floor(xv_hpx))
+y_frac=1.-(yv_hpx-Floor(yv_hpx))
+IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,degpix=obs.degpix,freq=obs.freq_center) $
```
nicholebarry, line 55: Please add a general comment here that this is the Jacobian for your FFT below

simulation/eor_bubble_sim.pro
```diff
@@ -29,23 +27,83 @@ print, "Npix_selected: ", npix_sel
 
 ; Limit the range of frequencies in the uvf cube to the range of the obs
 freq_arr = (*obs.baseline_info).freq
-lim = minmax(freq_arr)
-freq_inds = where((freq_hpx GT lim[0]) and (freq_hpx LT lim[1]) )
-freq_hpx = freq_hpx[freq_inds]
+;lim = minmax(freq_arr)
+;freq_inds = where((freq_hpx GE lim[0]) and (freq_hpx LE lim[1]), count_inds )
```
nicholebarry, line 31: Go ahead and remove commented out blocks

simulation/eor_bubble_sim.pro
```diff
@@ -29,23 +27,83 @@ print, "Npix_selected: ", npix_sel
...
+; There isn't a standard way to do this strictly within IDL, so call healpy functions using IDL to Python bridge.
+if keyword_set(ltaper) THEN BEGIN
+    print, 'Applying l taper'
+    lmax_fhd = floor(180./obs.degpix)
+    lmax_hpx = 3*nside-1
+    if lmax_fhd GT lmax_hpx THEN BEGIN
+        print, "Input map resolution below FHD settings. lmax_fhd="+string(lmax_fhd)
+        lmax_fhd = lmax_hpx
+    ENDIF
+    hp = python.import('healpy')
```
nicholebarry, line 68: Please do another check to see if these functions exist for IDL because we currently don't have any python dependencies. If not, find out what the minimum healpy version is required for this to work and add a caveat to our dependencies section on the front-page readme, i.e. if generating visibilities from a EoR hd5 file using the simulation framework, then healpy version xx is required.

simulation/eor_bubble_sim.pro
```diff
@@ -29,23 +27,83 @@ print, "Npix_selected: ", npix_sel
...
+    mmax_hpx= lmax_hpx
+    almsize = floor(mmax_hpx * (2 * lmax_hpx + 1 - mmax_hpx) / 2.) + lmax_hpx + 1
+    almidx = lindgen(almsize)
+    lm = hp.Alm.getlm(lmax_hpx,almidx)
+    lmat = lm[0]
+    mmat = lm[1]
+    win_mat = 0.5*(1-tanh(ltaper*(lmat-lmax_fhd)))
+    for fi=0, nfreq_hpx-1 DO BEGIN
+        print, fi
```
nicholebarry, line 77: remove print statement, add a comment on the general purpose of the loop

simulation/test_eor_sim.pro
```diff
@@ -7,19 +7,23 @@ FUNCTION test_eor_sim
 
 ; Visibility degrid requires -- obs, psf, image_uv, and vis_weight
 
-instrument='paper'
+instrument='mwa'
```
nicholebarry, line 10: Please move this file to the pipeline_scripts/FHD_IDL_wrappers repo

simulation/vis_simulate.pro
```diff
@@ -51,7 +51,8 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
       ENDELSE
 
     endfor
-    IF ~Keyword_Set(no_save) THEN save, file=init_beam_filepath, beam2_xx_image, beam2_yy_image, obs
+    test = strpos(init_beam_filepath, '_000_')
```
nicholebarry, line 54: I'm not sure what the purpose of this test is and if it's something generalised or specific to your testing.

simulation/vis_simulate.pro
```diff
@@ -262,20 +263,17 @@ FUNCTION vis_simulate,obs,status_str,psf,params,jones,skymodel,file_path_fhd=fil
...
-          this_model_ptr=vis_source_model(skymodel,obs,status_str,psf,params,this_vis_weight_ptr,model_uv_arr=this_model_uv,$
-            timing=model_timing,silent=silent,error=error,_Extra=extra)
-        ;  model_timing=0.0
-    ; vis_source_model uses DFT, which works for point source models. Since the uvf cube was already FFTed, just run visibility_degrid.
-        ;  FOR pol_i=0,n_pol-1 DO BEGIN
-        ;      (*vis_model_arr[pol_i])[fi,*] = (*(visibility_degrid(*this_model_uv[pol_i],this_vis_weight_ptr[pol_i],obs,psf,params,timing=timing,silent=silent,$
-        ;          polarization=pol_i,_Extra=extra)))[fi,*]
-        ;       model_timing += timing
-        ;  ENDFOR
+           ;this_model_ptr=vis_source_model(skymodel,obs,status_str,psf,params,this_vis_weight_ptr,model_uv_arr=this_model_uv,$
```
nicholebarry, line 266: go ahead a remove commented out blocks

nicholebarry (commented):
Looks like a much better setup than currently is what being done. Would it be possible to see what the power spectra and healpix image looks like for an example EoR sim? I feel like you've done this already, but I can't seem to find them

simulation/eor_bubble_sim.pro
aelanman: It looks like IDL's healpix does provide a version of anafast which will compute a_lm, so it should be possible to do this without invoking the Python bridge. This would involve writing out a_lm components to a fits file and then reading them back in.

However, this was only a feature I added while trying to get the healpix binning right, and isn't essential to HEALPix-based EoR simulations. I could just cut it altogether.

simulation/vis_simulate.pro
aelanman: I think this was a temporary measure I added because I was running hundreds of simulations and they kept saving the same beam image (and I only wanted one copy). I'll just delete it.

simulation/test_eor_sim.pro
aelanman: I've opened a PR on pipeline_scripts and removed the file from here.

aelanman:
@nicholebarry Thanks for the review, Nichole! I think I've addressed most of your comments.

A minor sticking point is that I no longer have access to the Brown cluster, and I no longer have a personal IDL license. So I can't run FHD/eppsilon anywhere at the moment. I don't think I have any old power spectrum plots on my laptop. There should be some in the cold-storage data that @jpober has on google drive.

simulation/eor_bubble_sim.pro
nicholebarry: Gotcha. How about comment it out and put a comment explaining what the purpose is so if anyone needs to repeat this for different binning etc, they don't have to reinvent the wheel

aelanman: Okay! I've commented out the block and left an instruction in dictionary.md to uncomment to re-enable it.

nicholebarry (approved):
Perfect, thanks for taking the time to see this through @aelanman 

@jpober when you have the time, I'd like to see a sample plot, but it's not critical

jpober:
I haven't been following this thread closely, but I don't think there's anyone with the institutional knowledge at Brown to run this bit of code / produce the plots you're looking for.  If someone can send us a walkthrough/script, we can get something run, but sounds like it's not pressing priority?

PR #125, opened 2018-04-26
Pull n_vis_cal from modified weights

nicholebarry:
Currently, n_vis_cal pulls from the weights pointer, which is purposely not modified by baseline cuts. However, this is false advertising: we want to include baseline cuts in the calculation of n_vis_cal. 

The change I pushed pulls it from a modified versions of the weights. I do this after a for loop over the polarizations, so it will grab the n_vis_cal from the YY pol (instead of the XX pol, which was the default before).

Please comment if you have an opinion.


bhazelton:
Can the xx pol ever have different numbers of visibilities than the yy pol, e.g. because of flagging? If so, I'd advocate for keeping both.

nicholebarry:
Tested, appears to work.

PR #127, opened 2018-05-09
Fix incorrect organization of eor simulation code

bhazelton:
`eor_sim` was in the obsolete folder but it's definitely not obsolete.
`eor_simulation` was not in the obsolete folder, but I think it should be because it's calling fhd_sim which is in the obsolete folder


bhazelton:
Is it ok to merge this @isullivan ?

isullivan:
Looks fine. Go ahead and merge.

PR #135, opened 2018-09-11
Layout bug fix

isullivan:
This re-enables creating the layout metadata structure, which was commented out last year by accident. 

It also replaces the new list datatype with older string arrays. This is done both for compatibility with IDL 7, and also because the `keyword_list.count('string')` syntax appears to be system dependent, and raises an error on my machine.


bhazelton:
Actually, there's one more field I'd like the code to check for. I'll add it and push it to this branch.

bhazelton:
@isullivan there are a couple of columns in the antenna table that are optional, but can matter for some analysis codes (namely `diameter` and `beamfwhm`). I can test whether they're present and only add them to the structure if they are, but I don't know how that plays with named structures -- don't they have to have fixed fields? How should I handle the fact that they may or may not be present?

bhazelton:
I pushed what I think is the right code, let me know if I should change anything @isullivan!

PR #139, opened 2018-11-16
Auto calibration

wenyang-li:
New keyword **decompose_auto** is added, default is off. If turned on, the amplitude of each antenna is decomposed into Sqrt_Auto(f)*Global_Bandpass(f)*Constant, and the cable reflection mode fitting is restricted to phase residual instead of gain residual because auto-correlations already have reflection modes in them.

I have attached my two runs of original calibration (cable-averaged bandpass, linear phase, quadratic amplitude, 150m cable reflection) minus new calibration (sqrt_auto-correlation times global_bandpass times const, linear phase, 150m cable reflection fitting for phase residual) for Phase I (15 obs, zenith) and Phase II (131 obs, 5 pointings). The difference ps for these two tests are attached.

Phase I, 15 obs, zenith:
cable_minus_auto_phase1_15obs_zenith

Phase II, 131 obs, 5 pointings:
cable_minus_auto_phase2_131obs_5pointings


wenyang-li:
I have the calibration_auto_fit=1 to run and made a ps comparison with the new calibration approach for the Phase I 15 obs.

auto_minus_autofit_phase1_15obs_zenith

I cannot say that the new approach is better than calibration_auto_fit, but I have two concerns for calibration_auto_fit: 1. the edge coarse channels (unflagged) are overestimated in amplitude, so the overall power gets reduced. 2. the coarse channels are bleeding to neighbor modes.

wenyang-li:
Just want to advertise this change a little bit. I made the following plot for calibration from Phase II  zenith pointing data. The calibration gets averaged over 30 min. Top: per cable type bandpass; Middle: per cable type bandpass with auto-correlations removed, you can see they overlap pretty well; Bottom: the averaged 'global bandpass' with auto-correlations removed. 

[bp.pdf](https://github.com/EoRImaging/FHD/files/2695498/bp.pdf)

I think there are two advantages: 1. Auto-correlations better captures antenna to antenna variation; 2. This plot shows that we can average over more samples, which gives more spectrally smooth bandpass.

bhazelton:
That's fascinating @wenyang-li! Can you explain what you're plotting in grey in the 3rd plot? I assume they're individual tile calibrations, but do they have the auto-correlations removed?

wenyang-li:
You are right @bhazelton. The gray ones are individual tile calibrations and there are 126 of them. They have autos removed.

bhazelton:
Ok, thanks! To show that the black lines are smoother than you'd get without removing the autos, can you also plot what the average looks like without the autos removed? I know we have plotted that before but it'd be nice to have it on the same scale to compare with your black line.

wenyang-li:
I mean the global bandpass with auto removed will be smoother than the cable averaged bandpass. If we use global bandpass without auto removed, we are losing structure variation between different cable types because we are averaging those real structures out, so I assume we are not going that way. I have also plotted the global bandpass without removing autos for comparison here.
[bp2.pdf](https://github.com/EoRImaging/FHD/files/2696454/bp2.pdf)

bhazelton:
Thanks! You're right of course, but I like this new plot as well because it shows how much tighter the grey lines are after you remove the auto correlations.

The shapes of the central parts of the coarse bands after auto removal is also very interesting. I wonder if that's coming in part from the differences in the shapes of the coarse bands in autos vs crosses -- as I recall the coarse band edges are higher in autos vs lower in the crosses. Maybe these shapes are compensating for the introduced auto shapes?

wenyang-li:
What you see in the global bandpass is actually the amplitude structure of antenna 0 (Tile 11), which has 90m cable. I calculated the scaled auto in this way: scaled_auto[i] = auto[i]/auto[0]. The global sky signal is changing over time so I want the signal part being divided out, and turns out the ratio is very consistent over a pointing. I did this because I wanted to do long term global bandpass average so I want to make them consistent between obs ids.

bhazelton:
Aahh, ok, that makes sense. I guess the only potential problem is if your reference antenna has a problem and you need to change your reference, but I think we can live with that.

wenyang-li:
I made a version of bandpass plots for Phase I GoldenSet, zenith pointing, where the pfb correction was on. This visualizes the cable reflection better.
[bp1.pdf](https://github.com/EoRImaging/FHD/files/2721606/bp1.pdf)

wenyang-li:
A few updates: 
1. The reference antenna I use for the 'scaled auto' will be the ref_antenna in cal struct, i.e., the same antenna for 0 phase reference. 
2. I have added 'mwa2_cable_length' and 'mwa2_cable_reflection_coefficient.txt' to instrument_config. We can use them for Phase II data bandpass calibration and cable reflection fitting. **The instrument keyword must be specified as 'mwa2', otherwise the cable information of Phase 1 will be mistakenly used.** 
3. The bandpass_plots was hard coded. It only allows 6 types of cables, otherwise it will throw an error. I have made it more flexible because phase 2 has only 4 types of cables. 
4. The amp_degree will be set to 0 if decompose_auto is on. This means if using autos to fit amplitudes, there will be only a constant scaling parameter for the amplitude residual fitting: Abs(g[i,f])=sqrt(auto[i,f])*global_bp[f]*const[i]

wenyang-li:
One more update: I changed the keyword a while ago to make it less confusing. 

To make it work, simply set auto_ratio_calibration=1.

nicholebarry:
This branch is now ready to be integrated. Significant changes needed to be made to harmonize this code with the current form of FHD. 

In particular, keywords were added to allow for some aspects of the auto-cal to be done during cross-cal (i.e. incoherent means and applying reflections to phases only). In addition, Li et al 2019 used redundant cal (I assume) to get the overall gain level, which isn't standard FHD. I've removed that dependence by allowing for polyfits of the gain to occur. 

Here is the 2D of one observation in the zenith pointing using this new version of auto-ratio-cal
fhd_nb_auto_ratio_uvf_1061317152_fullimg_dft_averemove_swbh_dencorr_2dkpower

Here is the 2D of the same observation using the master branch auto-cal-fit
fhd_nb_auto_fit_uvf_1061317152_fullimg_dft_averemove_swbh_dencorr_2dkpower

And here is the difference of the master auto cal and the auto ratio 
fhd_nb_auto_fullimg_dft_averemove_swbh_dencorr__fit_1061317152_minus_ratio_1061317152_2dkpower

There is good agreement with this new version of auto-ratio-cal and what Wenyang presented above. Extra keywords will be added to the documentation branch.

nicholebarry (approved):

PR #140, opened 2018-12-03
Add redundant claibration option using visibility covariances

isullivan:
The redundant calibration correction should only have an effect if `use_redundant_calibration=1` is set.
If set, it calculates the full covariance matrix between visibilities of the same frequency, and uses that to solve for a correction to the model visibilities given the input data and calculated gains.
The correction is calculated iteratively with the gains.
The covariance matrix is *very* slow to calculate, even using multiple optimizations. On my machine, regular calibration takes ~20 minutes, but including redundant calibration it takes ~2 days.


rlbyrne:
How often does the covariance matrix need to be calculated? Can we calculate one per frequency per instrument and read it in every time we want to calibrate that instrument?

isullivan:
In principle the covariance matrix depends on many details of the observation and should be different every time, just like the holographic mapping function: it will depend on the beam model (including the pointing) and the flagging. However, as long as we aren't changing the beam model over time we could probably construct a covariance matrix for each pointing, which could then be loaded and modified to accommodate the flagging.

PR #142, opened 2018-12-11
Add output message and settings file entry for cal convergence

isullivan:
If calibration exits because it hits the maximum number of iterations for a given frequency and polarization, it will now print a message to the log, e.g. `Calibration reached max iterations before converging for pol_i: 1 freq_i: 372`

In addition, the number of frequencies that converged is now stored in the `cal` structure and is visible in the settings file. If this does not match the number of frequencies that should have been used, you can investigate further by looking through the `cal.convergence` parameter.


rlbyrne:
Looks good to me! Thanks for this fix.

PR #145, opened 2019-01-24
kernel window option

nicholebarry:
Option to apply the kernel window (i.e. Tapered Gridded Estimator).

Things to add:
*Warning messages. This method requires a two step process that needs to be managed by the user. Warnings should be given to indicate this.
*Documentation. There should be sufficient documentation so that the user can run this correctly. 
*Better handling of the kernel window pointer. Currently, just the first antenna structure points to the kernel window memory location. And the kernel window is pulled from just one of the antennas during image power beam creation. This could lead to weird side-effects if something more interesting than a beam template for all antennas is used.


nicholebarry:
To do: force the dimension of the antenna kernel to increase when using the kernel window (maybe through debug_dim, or can I be smarter about it?)

nicholebarry:
Everything complete except better pointer handling and changing debug_dim to be a better keyword. These can be moved to issues on the master branch

Issue #59, opened 2016-09-06
tile_flag_list used differently depending on type


nicholebarry:
In vis_flag_basic.pro, lines 31-59 on current master branch, there are two different ways tile_flag_list is interpreted. If it is a string, it is interpreted as a tile name. If it is a integer, it is interpreted as a tile index.

This is confusing. We should keep consistent for whatever is chosen. There should also be an indication (as an output message) on what is assumed.

isullivan:
In that case we need to only allow the tile name as a string, since some instruments might have non-integer tile names. It can return an error if an integer or other type is supplied.

nicholebarry:
Sounds like a plan. I can add it to the dictionary.

PR #147, opened 2019-02-07
Remove int option for flagging tiles

nicholebarry:
Return a message specifying that tile_flag_list only accepts strings.

Satisfies https://github.com/EoRImaging/FHD/issues/59


nicholebarry:
is stale, already implemented

PR #150, opened 2019-02-15
Jones matrix normalization was off by 2

isullivan:


wenyang-li:
I have done the test for these 3 versions:

Before full-pol merge, which gives the correct answer: b1d1a49
After the full-pol merge, which gives right intensity for I and Q, but gives half intensity in xx and yy, so does the most recent master branch: d8cff8b
The version of this branch, which has this issue fixed: cf52997

The I, Q, U, V plots for these 3 versions is here:
full_pol_merge_bug

Issue #153, opened 2019-03-25
Images fail when max_baseline is small


rlbyrne:
FHD fails to make images when max_baseline is small. For example, for a run with max_baseline=50 the output images look like:
![1131713632_uniform_Dirty_I](https://user-images.githubusercontent.com/12840057/54898317-043b6c00-4e89-11e9-98e7-07b109bf0b88.png)
This is true for all dirty, model, and residual images with n_pol=2 or 4.

The visibilities themselves are zeros or NANs.

PR #155, opened 2019-03-25
bug fix: beam resolution mismatch

rlbyrne:
This pull request resolves issue #153, where imaging was failing when max_baseline was small.

Imaging failed because the Jones matrix was infinite. This came from line 69 in fhd_struct_init_jones.pro, where power_zenith was being set to 0. In line 70, the Jones matrix is divided by power_zenith, leading to infinite values and NANs for the visibilities. In line 69 power_zenith_beam has a resolution of 2048 and obs_temp.zenx and obs_temp.zeny were each ~128. After the fix obs_temp.zenx and obs_temp.zeny are ~1024 and the interpolation in line 69 gives a non-zero value.

### DIFF
--- a/fhd_core/polarization/fhd_struct_init_jones.pro
+++ b/fhd_core/polarization/fhd_struct_init_jones.pro
@@ -55,7 +55,7 @@ hour_angle = hour_angle mod 360.
 apply_astrometry, obs, ra_arr=obs.obsra, dec_arr=obs.lat, x_arr=x_t, y_arr=y_t, /ad2xy
 apply_astrometry, obs, dec_arr=lat, x_arr=x_t, y_arr=y_t, /xy2ad, /ignore_refraction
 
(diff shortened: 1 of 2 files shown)


isullivan:
This fix looks fine, but I am worried that it is just covering up a bug in `fhd_struct_update_obs.pro`. I think that should default to values from the input obs structure if `degpix` and `kbinsize` are not supplied, instead of guessing values. I'm happy to make those changes, which are easy, if you like.

rlbyrne:
Ok! That sounds good, thanks.

isullivan:
@rlbyrne Please check my changes carefully, and run some test data. I'm worried that we might have bugs in other places where it happened to work because of this bug, but will be broken now that it's fixed.

rlbyrne:
I did a 4-pol test run and everything checks out. Are we ready to merge?

isullivan:
Since everything checked out, I went ahead and merged.

Issue #154, opened 2019-03-25
Imaging artifact in Stokes Q and U when long baselines are excluded


rlbyrne:
Stokes Q and U images have a discontinuity that is only visible when long baselines are excluded from imaging. 

Here is an example. For this run, baseline_threshold=-50, meaning that only baselines shorter than 50 wavelengths were used:
Stokes Q Dirty:
![1131713632_uniform_Dirty_Q](https://user-images.githubusercontent.com/12840057/54898620-38fbf300-4e8a-11e9-9e02-0727effb9db5.png)
Stokes U Dirty:
![1131713632_uniform_Dirty_U](https://user-images.githubusercontent.com/12840057/54898628-3f8a6a80-4e8a-11e9-8192-7676e77a7a74.png)

For reference, here is the same observation with identical calibration, now imaging with all baselines. Here you can't see the artifact:
Stokes Q Dirty:
![1131713632_uniform_Dirty_Q (1)](https://user-images.githubusercontent.com/12840057/54898685-7f515200-4e8a-11e9-90db-5e41fd2a8fb7.png)
Stokes U Dirty:
![1131713632_uniform_Dirty_U (1)](https://user-images.githubusercontent.com/12840057/54898690-82e4d900-4e8a-11e9-8d24-bad50ffef560.png)

This artifact often appears at the center of the observation but not always. In this observation it's in the top part:
Stokes Q Dirty:
![1130773144_uniform_Dirty_Q](https://user-images.githubusercontent.com/12840057/54898927-4bc2f780-4e8b-11e9-95f4-c1ecaef2c364.png)
Stokes U Dirty:
![1130773144_uniform_Dirty_U](https://user-images.githubusercontent.com/12840057/54898940-59787d00-4e8b-11e9-9ba0-e6a2720dad84.png)

isullivan:
It looks like the discontinuity is always located at zenith. Is that correct? I suspect it is present in the dirty images as well, but it's hard to see since the color scale is ~10x larger. If you save the dirty Stokes images as fits files and zoom in at zenith and maybe smooth them a little, can you see this feature?

rlbyrne:
@isullivan These are the dirty images. Do you mean the all-baselines images?

rlbyrne:
Here I've plotted the residual fits images around the imaging artifact with all baselines and with the baseline cut. I really can't say if the imaging artifact just isn't there when we use all baselines or if it's just masked by the small-scale structure.
[...]

PR #156, opened 2019-04-03
Stokes basis coordinates

isullivan:
This addresses issue #154 by rotating the basis vectors by the parallactic angle (per pixel) when constructing the Jones matrix. If it works as intended, it should leave unpolarized sources unchanged, and fix the artifacts we are seeing for polarized sources. It should not affect the Jones matrix used to construct the gridding/degridding kernels.


rlbyrne:
I just confirmed that these changes do not affect the power spectrum at all when the model is unpolarized. The branch is ready to be merged.

PR #158, opened 2019-05-13
Make the Jones matrix double precision and correct a Stokes conversion bug

rlbyrne:
This pull request changes the Jones matrix to double precision, from float. It also includes a bug fix for the Stokes conversion.

These changes were motivated by work done on the Stokes_basis_coordinates branch. We found that Stokes I source modeling depended on the coordinate system. This dependency was due to machine precision errors and disappeared when we began using a double precision Jones matrix. 

Here is a power spectrum difference plot of master minus double_precision_beam:
fhd_rlb_averemove_swbh_dencorr__testing_before_stokes_convert_fix_Apr2019_1061316296_minus_test_double_precision_beam_branch_May2019_1061316296_2dkpower

### DIFF
--- a/fhd_core/polarization/stokes_cnv.pro
+++ b/fhd_core/polarization/stokes_cnv.pro
@@ -49,6 +49,10 @@ ENDELSE
 n_pix=N_Elements(inds)
 IF Keyword_Set(inverse) THEN p_use=p_map ELSE p_use=p_corr
 
+stokes_mat_term1=[1,1,1,complex(0,1)]
+stokes_mat_term2=[1,-1,1,-complex(0,1)]
+stokes_inv_term1=[.5,.5,.5,.5]
+stokes_inv_term2=[.5,-.5,-complex(0,.5),complex(0,.5)]
 stokes_list1=[0,0,2,2]
 stokes_list2=[1,1,3,3]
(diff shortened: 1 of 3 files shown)


fhd_core/polarization/stokes_cnv.pro
```diff
@@ -172,6 +173,7 @@ ENDIF ELSE BEGIN ;else case is array of images
     ;stokes I can have proper inverse-variance weighting. (not used!)
     ; All other polarizations need to be converted to 'true sky' frame before they can be added
     IF Keyword_Set(inverse) THEN BEGIN ;Stokes -> instrumental
+        sign=[1,-1,-complex(0,1),complex(0,1)]
```
isullivan: For consistency, let's make this sign1 = [1, 1, 1, 1] and sign2=[1,-1,-complex(0,1),complex(0,1)], and use both on line 178 below.

fhd_core/polarization/stokes_cnv.pro
```diff
@@ -49,6 +49,10 @@ ENDELSE
 n_pix=N_Elements(inds)
 IF Keyword_Set(inverse) THEN p_use=p_map ELSE p_use=p_corr
 
+stokes_mat_term1=[1,1,1,complex(0,1)]
```
isullivan: Please add a comment that this is a subset of the full 4x4 matrices, with the zero terms omitted

PR #160, opened 2019-06-04
Bug fix: uvfits headers can have 10+ parameters

isullivan:
The crash using the new uvfits headers with LST added was because the number of parameters increased from 9 to 11. The header interpreter formatted the "PTYPE" keywords with only a single integer, so parameters past 9 could not be read. This removes that limitation.


wenyang-li:
I have just tried to read the uvfits with 11 parameters and it went through. I think this is fixed.

PR #165, opened 2019-07-13
Add error handling for when no sources found

mwilensky768:
This is a small change that just gives a slightly more informative error message in stokes_cnv when all sources in the catalog are below the horizon.

### DIFF
--- a/fhd_core/polarization/stokes_cnv.pro
+++ b/fhd_core/polarization/stokes_cnv.pro
@@ -6,6 +6,8 @@ FUNCTION stokes_cnv,image_arr,jones,obs,beam_arr=beam_arr,inverse=inverse,square
 ;;Note that "image_arr" can actually be a 2D image, a vector of values, or a source_list structure. 
 ;requires the jones structure from fhd_struct_init_jones.pro
 
+ON_ERROR, 2
+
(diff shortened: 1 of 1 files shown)


fhd_core/polarization/stokes_cnv.pro
```diff
@@ -117,6 +117,7 @@ IF type EQ 8 THEN BEGIN ;check if a source list structure is supplied
     ind_arr[inds]=Lindgen(N_Elements(inds))
     p_ind=ind_arr[sx,sy]
     s_use=where(p_ind GE 0,ns)
+    IF ns EQ 0 THEN message, "Error: ns=0, probably no sources above the horizon"
```
isullivan: If you're going to use `message` you also should define what it should do when an error is raised. The default is to halt at the line that raised the error, which is equivalent to setting `ON_ERROR, 0` near the beginning of the function. In this case, I think either `ON_ERROR, 1` (which halts execution and returns to the main program level) or `ON_ERROR, 2` (which halts execution and returns to the line in the function that called the one with the error) would be preferable. 
If you add `ON_ERROR, 1` or `ON_ERROR, 2`, that should go just above line 9 in this file.

mwilensky768:
Thanks @isullivan. I opted to call to level 2 on error by putting ON_ERROR above line 9, as you suggested.

PR #169, opened 2019-07-25
Use healpix_recalculate keyword

rlbyrne:
The keyword healpix_recalculate is mentioned in the dictionary but doesn't actually do anything. This pull requests implements a fix to actually recalculate the Healpix cubes when healpix_recalculate is set.


rlbyrne:
Got it. Should we just snapshot_recalculate then? I wasn't sure what snapshot_recalculate did so that's why I was looking at healpix_recalculate. Maybe this could just be resolved with a dictionary update.

isullivan:
No, I think healpix_recalculate is more descriptive, so I would suggest we just change the code to match the documentation.

rlbyrne:
I renamed snapshot_recalculate. Does this look okay now?

Issue #171, opened 2019-07-31
Debug option for identical X and Y beams


rlbyrne:
It would be useful for debugging to have a keyword that makes the X and Y beams equal, similar to debug_flip (which swaps X and Y beams). I currently can't figure out how to make the beams equal.

rlbyrne:
I'm closing this issue for the time being because it's not clear it's possible (see #174)

PR #174, opened 2019-08-06
Create keyword debug_equal_beams

rlbyrne:
This PR creates a new keyword that sets the Y beam equal to the X beam for debugging (issue #171). The branch has been tested and works.


rlbyrne:
It turns out we can't make the beams (and Jones components for X and Y) equal because that makes the Jones matrix singular. If we want to include this option in FHD we need to go back to the drawing board about how to implement it.

PR #175, opened 2019-08-07
Rework the crosspol phase calculation

rlbyrne:
The crosspol phase is the overall phase offset between the X and Y dipoles. This phase is degenerate in typical FHD calibration, which only uses XX and YY visibilities. Previously, the crosspol phase was calculated by minimizing the average "pseudo-V". This PR replaces that calculation with a more rigorous and well-motivated calculation by calibrating to the model XY and YX visibilities. These changes have been tested and have little impact on the Stokes images.


fhd_core/calibration/vis_calibrate_crosspol_phase.pro
```diff
@@ -13,24 +13,33 @@ n_baselines=obs.nbaselines
...
+print,"Phase fit between X and Y antenna polarizations:", cross_phase
 
 cal.cross_phase = cross_phase
-*(cal.gain[0]) *= Exp(-icomp * cross_phase / 2.0)
-*(cal.gain[1]) *= Exp(icomp * cross_phase / 2.0)
+*(cal.gain[0]) *= Exp(-icomp*cross_phase/2.0)
+*(cal.gain[1]) *= Exp(icomp*cross_phase/2.0)
+
+*vis_ptr[2] *= Exp(icomp*cross_phase)
+*vis_ptr[3] *= Exp(-icomp*cross_phase)
```
isullivan: I don't think I like modifying the visibilities outside of `vis_calibration_apply`. Since we store `cal.cross_phase`, I think the correct way would be to not modify the `cal.gain` or `vis_ptr` here at all, but include new code in `vis_calibration_apply` along the lines of:
```
IF n_pol EQ 4 THEN BEGIN
*vis_ptr[2] *= Exp(icomp*cal.cross_phase)
*vis_ptr[3] *= Exp(-icomp*cal.cross_phase)
ENDIF
```

rlbyrne: That makes complete sense. The only problem is that in this implementation vis_calibrate_crosspol_phase uses the calibrated visibilities (calibrated up to the crosspol phase), not the raw visibilities. How would you recommend doing this?

isullivan: I'm going to push a commit with my suggested changes. Please look those over and verify that those do what you want.

rlbyrne: I've fixed the bugs in those changes. The new code has been tested and returns the expected result. I'm going to go ahead and merge.

PR #182, opened 2020-01-07
Use per-pol image normalization for all image filters

rlbyrne:
Using a single normalization across polarizations caused catastrophic errors in imaging when "optimal weighting" (FFT filter "fiter_uv_weighted") is used. PR #181 implemented per-pol image normalization for that weighting scheme only. This PR implements per-pol normalization for all weighting schemes, not just optimal.

PR #185, opened 2020-01-16
Correct diffuse normalization

rlbyrne:
This PR corrects the normalization of input diffuse maps (which can now be input in units of Jy/sr) and output images (which are now output in Jy/sr). It does proper diffuse normalization when optimal weighting is used (image_filter_fn='filter_uv_weighted'). It also makes a change to output model Healpix fits files when the keyword write_healpix_fits is set.


fhd_core/HEALPix/healpix_interpolate.pro
```diff
@@ -49,8 +49,8 @@ image_mask[Min(Floor(xv_hpx)):Max(Ceil(xv_hpx)),Min(Floor(yv_hpx)):Max(Ceil(yv_h
 x_frac=1.-(xv_hpx-Floor(xv_hpx))
 y_frac=1.-(yv_hpx-Floor(yv_hpx))
 IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,nside=nside,freq=obs.freq_center) $
+    ELSE IF keyword_set(from_jy_per_sr) THEN pixel_area_cnv=(obs.degpix*!DtoR)^2. $ ; (steradian/new pixel)
     ELSE pixel_area_cnv=((obs.degpix*!DtoR)^2.)/(4.*!Pi/n_hpx) ; (steradian/new pixel)/(steradian/old pixel)
```
isullivan: I would suggest either using a CASE statement here (which allows an ELSE), or indent the final ELSE statement here one more level.

fhd_core/source_modeling/fhd_diffuse_model.pro
```diff
@@ -52,7 +52,8 @@ IF size(diffuse_spectral_index,/type) EQ 10 THEN BEGIN ;check if pointer type
     print,"A spectral index is defined in the saved diffuse model, but this is not yet supported!"
 ENDIF ;case of specifying a single scalar to be applied to the entire diffuse model is treated AFTER building the model in instrumental polarization
 
-model_stokes_arr=healpix_interpolate(model_hpx_arr,obs,nside=nside,hpx_inds=hpx_inds,from_kelvin=diffuse_units_kelvin,coord_sys=coord_use)
+if ~keyword_set(diffuse_units_kelvin) then from_jy_per_sr=1
```
isullivan: Since this is setting the default value for these parameters I would move it to the top of the file, maybe right under `IF Keyword_Set(flatten_spectrum)`...

Also, this leaves `from_jy_per_sr` undefined if `diffuse_units_kelvin` is set. If we later set default values for these parameters that could cause a bug, so I would recommend being explicit. Something like:
```
IF N_Elements(diffuse_units_kelvin) EQ 0 THEN diffuse_units_kelvin = 0
from_jy_per_sr = Keyword_Set(diffuse_units_kelvin) ? 0:1
```

fhd_output/fhd_quickview.pro
```diff
@@ -276,13 +276,14 @@ ENDFOR
...
     dimension=dimension,restored_beam_width=restored_beam_width,_Extra=extra)
 
-; renormalize based on weights
-renorm_factor = get_image_renormalization(obs_out,weights_arr=weights_arr,beam_base=beam_base_out,filter_arr=filter_arr,$
-  image_filter_fn=image_filter_fn,pad_uv_image=pad_uv_image,degpix=degpix,/antialias)
-for pol_i=0,n_pol-1 do begin
-  *instr_dirty_arr[pol_i]*=renorm_factor[pol_i]
-  IF model_flag THEN *instr_model_arr[pol_i]*=renorm_factor[pol_i]
-endfor
+IF keyword_set(image_filter_fn) THEN IF image_filter_fn NE 'filter_uv_weighted' THEN BEGIN ; renormalize based on weights
```
isullivan: We need to set the default for `image_filter_fn` to `filter_uv_uniform` at the start of this file. Otherwise it just gets set to that inside `dirty_image_generate` and that could be surprising. That means that you can treat it as guaranteed to be defined and can skip the `IF keyword_set(image_filter_fn)`... test here.

Then, you should modify `get_image_renormalization` so that it returns the correct normalization regardless of the filter function. For example, it should know to return `1` for `filter_uv_weighted` (right at the start!), and should divide by `(degpix*!DtoR)^2. in all other cases. I think other code relies on `get_image_renormalization`, so you should add a new keyword to convert to Jy/sr and set that when you call it here.

isullivan: After you've made the above changes, you shouldn't need an `IF...` statement here at all.

fhd_output/fhd_quickview.pro
```diff
@@ -276,13 +276,14 @@ ENDFOR
...
+    renorm_factor = get_image_renormalization(obs_out,weights_arr=weights_arr,beam_base=beam_base_out,filter_arr=filter_arr,$
+      image_filter_fn=image_filter_fn,pad_uv_image=pad_uv_image,degpix=degpix,/antialias)
+    for pol_i=0,n_pol-1 do begin
+      *instr_dirty_arr[pol_i]*=renorm_factor[pol_i]/(degpix*!DtoR)^2.
+      IF model_flag THEN *instr_model_arr[pol_i]*=renorm_factor[pol_i]/(degpix*!DtoR)^2.
```
isullivan: I'm surprised I'm only catching this now, but shouldn't we also renormalize `instr_sources` if `source_flag` is set, and `instr_rings` if `ring_radius` is also set?

fhd_core/HEALPix/healpix_interpolate.pro
rlbyrne: I indented the final ELSE statement.

fhd_core/source_modeling/fhd_diffuse_model.pro
rlbyrne: I believe I resolved this but it could use a final review

fhd_output/fhd_quickview.pro
rlbyrne: I also think this should be resolved now but could use another review

fhd_output/fhd_quickview.pro
rlbyrne: You mean to get consistent scaling for the restored images? I agree in that case. This should be fixed.

rlbyrne:
You're absolutely right that I should check with users about the normalization change. I'll post about it Slack - I don't know if there're better ways to reach FHD users.

More generally, it doesn't seem that the images were actually normalized in Jy/beam before. I think they were normalized in Jy/pixel. We can chat about this more if you want.

fhd_output/fhd_quickview.pro
rlbyrne: This has been fixed and tested and is ready for a code review.

fhd_output/fhd_quickview.pro
```diff
@@ -28,6 +28,7 @@ IF N_Elements(beam_threshold) EQ 0 THEN beam_threshold=0.05
 IF N_Elements(beam_output_threshold) EQ 0 THEN beam_output_threshold=beam_threshold/2.
 IF N_Elements(image_mask_horizon) EQ 0 THEN image_mask_horizon=1
 IF status_str.fhd EQ 0 THEN model_recalculate=0
+IF ~keyword_set(image_filter_fn) THEN image_filter_fn='filter_uv_uniform'
```
isullivan: Minor change: please also print a warning message if `image_filter_fn` was not defined and it uses the default.

fhd_output/fhd_quickview.pro
```diff
@@ -276,12 +277,13 @@ ENDFOR
...
     dimension=dimension,restored_beam_width=restored_beam_width,_Extra=extra)
 
-; renormalize based on weights
 renorm_factor = get_image_renormalization(obs_out,weights_arr=weights_arr,beam_base=beam_base_out,filter_arr=filter_arr,$
   image_filter_fn=image_filter_fn,pad_uv_image=pad_uv_image,degpix=degpix,/antialias)
 for pol_i=0,n_pol-1 do begin
   *instr_dirty_arr[pol_i]*=renorm_factor[pol_i]
   IF model_flag THEN *instr_model_arr[pol_i]*=renorm_factor[pol_i]
+  IF source_flag THEN *instr_sources[pol_i]*=renorm_factor[pol_i]
+  IF Keyword_Set(ring_radius) THEN *instr_rings[pol_i]*=renorm_factor[pol_i]
```
isullivan: `renorm_factor` is now either 1 for `filter_uv_weighted` or set to the average calculated from all polarizations. Doesn't that mean it never actually is different for the different polarizations? I think we need to decide whether it *should* always be the same for all polarizations and make `renorm_factor` a float (instead of an array), or remove the bit at the end of `get_image_renormalization` that sets each element equal to the mean value.

That investigation could be done as a new branch and PR, though, if you'd like to get these changes merged first.

fhd_output/get_image_renormalization.pro
```diff
@@ -15,8 +16,9 @@ FOR pol_i=0,n_pol-1 DO BEGIN
         weights=*weights_arr[pol_i],pad_uv_image=pad_uv_image,image_filter_fn=image_filter_fn,$
         filter=filter_arr[pol_i],antialias=antialias,beam_ptr=beam_base[pol_i],_Extra=extra))[dimension/2.,elements/2.]
     renorm_factor[pol_i]*=((*beam_base[pol_i])[obs.obsx,obs.obsy])^2.
+    renorm_factor[pol_i]/=(degpix*!DtoR)^2. ; Convert from Jy/pixel to Jy/sr
 ENDFOR
-IF keyword_set(image_filter_fn) THEN IF image_filter_fn NE 'filter_uv_weighted' THEN renorm_factor[*]=mean(renorm_factor)
+renorm_factor[*]=mean(renorm_factor)
 
 ;pix_area=beam_width_calculate(obs,min_restored_beam_width=0,/area)
 ;renorm_factor/=pix_area
```
isullivan: It's time to delete this commented-out code.

fhd_output/fhd_quickview.pro
rlbyrne: I do think that is should be per-polarization for all image filters. PR #182 does that (or did that, it needs to be updated in light of recent changes). I think it would make sense to do that as a separate PR but we can also implement it here.

isullivan: If that is coming soon, just leave this code as it is. That will make it easier to isolate the changes if something looks broken later.

PR #196, opened 2020-02-21
Documentation

nicholebarry:
It's time!


assumptions.md
```diff
@@ -2,15 +2,13 @@
 There are some assumptions about the data that are inherent to FHD and its mathematical basis. Documented below are these assumptions and their consequences. <br />
 
 ## Time-based <br />
-
-**Length of observations** <br />
-PSFs are calculated only once per snapshot (ObsID). Sources will move through the beam over the course of a given snapshot, but the assumption is that the sky is stationary. This leads to small discontinuities visible in waterfall plots. Choose snapshots shorter than about 2 minutes to minimize the discontinuity between adjacent snapshots. Also, phase to the zenith at the center of the snapshot.
+PSFs are calculated only once per observation. Sources will move through the beam over the course of a given snapshot, but the assumption is that the sky is stationary. This leads to small discontinuities visible in waterfall plots. Choose snapshots shorter than about 2 minutes to minimize the discontinuity between adjacent snapshots. Also, phase to the zenith at the center of the snapshot.
```
bhazelton: Sorry, just looking through this and I think this isn't quite right. I think the primary beam is only calculated once per observation, but the PSF (which depends on the baseline uvws is calculated separately for each integration inside the observation. So I think the text is correct except that the first word should be Primary beam rather than PSF.

assumptions.md
```diff
@@ -2,15 +2,13 @@
...
 ## W-projection <br />
-Needs updated by Ian/Bryna/Miguel <br />
+FHD does not employ any w-projection or w-stacking algorithms. Integration from FHD needs to be performed in image space to avoid decorrelation as the sky moves. In addition, this means that altitude variations in elements are not properly accounted for, and thus the array must be relatively coplanar.
 
 ## Source catalog <br />
-Spectral slope from one catalog, would be great if Patti could update <br />
+Sources are currently assumed to be point sources, and extended sources and diffuse emission are assumed to be point-source component models. The spectral slope is assumed to be a power law, with one general spectral slope for all sources in the catalog (future implementations will include gaussian/shapelet models and varied spectral slope).
 
 ## Cosmological <br />
 Maybe Bryna can illuminate. <br />
```
bhazelton: I don't think this section belongs in FHD. FHD doesn't use cosmological info, eppsilon does.

PR #201, opened 2020-02-25
Replace keyword max_sources with max_deconvolution_components

### LINKED ISSUE #188: Rename keyword "max_sources" to "max_deconvolution_sources"
The keyword "max_sources" refers to deconvolution only. I would be helpful if that keyword were renamed "max_deconvolution_sources" to avoid confusion.

rlbyrne:
Closes #188. We should update the pipeline_scripts repo before merging.

### DIFF
--- a/Observations/general_obs.pro
+++ b/Observations/general_obs.pro
@@ -2,7 +2,7 @@ PRO general_obs,cleanup=cleanup,recalculate_all=recalculate_all,export_images=ex
     mapfn_recalculate=mapfn_recalculate,grid_recalculate=grid_recalculate,snapshot_recalculate=snapshot_recalculate,deconvolve=deconvolve,$
     image_filter_fn=image_filter_fn,data_directory=data_directory,output_directory=output_directory,n_pol=n_pol,precess=precess,$
     vis_file_list=vis_file_list,fhd_file_list=fhd_file_list,healpix_path=healpix_path,catalog_file_path=catalog_file_path,$
-    complex_beam=complex_beam,pad_uv_image=pad_uv_image,max_deconvolution_components=max_deconvolution_components,$
+    complex_beam=complex_beam,pad_uv_image=pad_uv_image,$
     update_file_list=update_file_list,combine_healpix=combine_healpix,start_fi=start_fi,end_fi=end_fi,skip_fi=skip_fi,flag_visibilities=flag_visibilities,$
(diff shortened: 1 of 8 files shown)


rlbyrne:
This PR has been tested and works well. It should be merged in conjunction with https://github.com/EoRImaging/pipeline_scripts/pull/7.

examples.md
```diff
@@ -61,7 +61,7 @@ Fast Holographic Deconvolution: A New Technique for Precision Radio Interferomet
 Deconvolution settings   
 ~~~
 deconvolve = 1 
-max_sources = 200000
+max_deconvolution_components = 200000
```
isullivan, line 64: I think the value used for the example should match the default value in `general_obs`. That is currently 10,000, though the dictionary says 100,000 and we have 200,000 here. I suggest changing them all to match the value in the dictionary.

isullivan: I see that it is also set in `fhd_init`, though that will always be ignored since this default is set first. We must remove default values from `general_obs` that we actually want set somewhere else.

Observations/general_obs.pro
```diff
@@ -102,7 +102,7 @@ IF N_Elements(complex_beam) EQ 0 THEN complex_beam=1
 IF N_Elements(n_pol) EQ 0 THEN n_pol=2
 IF N_Elements(precess) EQ 0 THEN precess=0 ;set to 1 ONLY for X16 PXX scans (i.e. Drift_X16.pro)
 IF N_Elements(gain_factor) EQ 0 THEN gain_factor=0.15
-IF N_Elements(max_sources) EQ 0 THEN max_sources=10000. ;maximum total number of source components to fit
+IF N_Elements(max_deconvolution_components) EQ 0 THEN max_deconvolution_components=10000. ;maximum total number of source components to fit
```
isullivan, line 105: `max_deconvolution_components`, `gain_factor`, `add_threshold`, and `independent_fit` should all be removed from `general_obs`, since they are set in `fhd_init`, and the default values here will override those and cause confusion. You will need to remove them everywhere in `general_obs`.

isullivan: This is probably a separate issue, but we should also look through the rest of `general_obs` and make sure we're not setting any other defaults that we don't want to.

rlbyrne: Done!

PR #205, opened 2020-03-05
Fix undefined variable bug in beam setup

### LINKED ISSUE #204: Beam setup error
Recent changes seem to be causing some problems in the MWA beam setup. I'm not entirely up to date on what these changes do, but when I run FHD with MWA beams it fails on this line:

https://github.com/EoRImaging/FHD/blob/797d9fc2219e8188a7c1c872e0bd16c229aee8ae/fhd_core/beam_modeling/beam_power.pro#L34

The variable `di_uv_use` is not defined in this procedure, nor can I see where it gets calculated. It loks like this was changed last November, on commit bd90cac50.

rlbyrne:
Closes #204


aelanman:
Thanks Ruby!

PR #216, opened 2020-06-01
Reduce mem

nicholebarry:
To reduce memory load, only store values of the zero-padded beam image which are not zero. Subsequently, store their array index so that zero-padded beam images can be made on the fly during degridding/gridding.

This reduces the memory load significantly, up to 50% when 384 freq-dependent beams are used.

One additional commit: force the gaussian source models to use the true dft versus the dft approximation. @isullivan, can you confirm whether or not the gaussian source models can use the dft approximation? The gaussian source models are made be multiplying a point source uv-plane with a gaussian envelope, a la [Line et al. 2020](https://arxiv.org/abs/2005.09316) Eq 11.

### DIFF
--- a/fhd_core/deconvolution/source_detection/source_comp_init.pro
+++ b/fhd_core/deconvolution/source_detection/source_comp_init.pro
@@ -13,13 +13,8 @@ beam_struct={beam,xx:0.,yy:0.,xy:Complex(0.),yx:Complex(0.),I:0.,Q:0.,U:0.,V:0.}
 shape_struct={x:0.,y:0.,angle:0.} ;gets used if keyword gaussian_source_models is set
 ;flux order is 0-3: xx, yy, xy, yx in apparent brightness; 4-7: I, Q, U, V in sky brightness
 ;flag type codes are 0: no flag, 1: low confidence 2: sidelobe contamination
-IF keyword_set(shape) THEN BEGIN
-    struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
-        flux:flux_struct,shape:shape_struct,beam:beam_struct}
-ENDIF ELSE BEGIN
-    struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
-        flux:flux_struct,beam:beam_struct}
-ENDELSE
+struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
+    flux:flux_struct,shape:shape_struct,beam:beam_struct}
 source_comp_new=Replicate(struct_base,n_sources>1)
 
 IF Keyword_Set(xvals) THEN source_comp_new.x=xvals
--- a/fhd_core/deconvolution/source_detection/source_comp_init.pro
+++ b/fhd_core/deconvolution/source_detection/source_comp_init.pro
@@ -13,8 +13,13 @@ beam_struct={beam,xx:0.,yy:0.,xy:Complex(0.),yx:Complex(0.),I:0.,Q:0.,U:0.,V:0.}
 shape_struct={x:0.,y:0.,angle:0.} ;gets used if keyword gaussian_source_models is set
 ;flux order is 0-3: xx, yy, xy, yx in apparent brightness; 4-7: I, Q, U, V in sky brightness
 ;flag type codes are 0: no flag, 1: low confidence 2: sidelobe contamination
-struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
-    flux:flux_struct,shape:shape_struct,beam:beam_struct}
+IF keyword_set(shape) THEN BEGIN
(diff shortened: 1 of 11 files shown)


nicholebarry:
ping @isullivan

fhd_core/deconvolution/source_detection/source_comp_init.pro
```diff
@@ -13,8 +13,13 @@ beam_struct={beam,xx:0.,yy:0.,xy:Complex(0.),yx:Complex(0.),I:0.,Q:0.,U:0.,V:0.}
...
 ;flag type codes are 0: no flag, 1: low confidence 2: sidelobe contamination
-struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
-    flux:flux_struct,shape:shape_struct,beam:beam_struct}
+IF keyword_set(shape) THEN BEGIN
+    struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
+        flux:flux_struct,shape:shape_struct,beam:beam_struct}
+ENDIF ELSE BEGIN
+    struct_base={id:-1L,x:0.,y:0.,ra:0.,dec:0.,ston:0.,freq:100.,alpha:0.,gain:1.,flag:0,extend:Ptr_new(),$
+        flux:flux_struct,beam:beam_struct}
+ENDELSE
```
isullivan, line 22: I think some code might break if the source structure might have different components for different uses. It might be okay to just always include `shape`

fhd_core/source_modeling/source_dft_multi.pro
```diff
@@ -129,14 +129,19 @@ IF Keyword_Set(n_spectral) THEN BEGIN
...
     IF Keyword_Set(dft_threshold) THEN BEGIN
         model_uv_arr=Ptrarr(n_pol,n_spectral+1)
-        edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix),$
-            n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
+        if ~keyword_set(gaussian_source_models) then begin
+            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix),$
+                n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
+        endif else begin
+            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix) OR $
+                gaussian_x,n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
```
isullivan, line 137: Won't this flag everything except a column of pixels along gaussian_x=0?

I think the correct way to do this is to make a `dft_threshold_use` variable (maybe with a better name) that is forced to 0 if `gaussian_source_models` is set.

fhd_core/source_modeling/source_dft_multi.pro
```diff
@@ -196,14 +201,19 @@ ENDIF ELSE BEGIN
     ENDIF
     IF Keyword_Set(dft_threshold) THEN BEGIN
         model_uv_arr=Ptrarr(n_pol)
-        edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix),$
-            n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
+        if ~keyword_set(gaussian_source_models) then begin
+            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix),$
+                n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
+        endif else begin
```
isullivan, line 207: Same as above.

instrument_config/mwa_beam_setup_gain.pro
```diff
@@ -86,12 +87,11 @@ CASE beam_model_version OF
 ;        IF n_zenith GT 0 THEN $
 ;            FOR p_i=0,n_ant_pol-1 DO FOR p_j=0,n_ant_pol-1 DO FOR freq_i=0L,nfreq_bin-1 DO (*Jmat_interp[p_i,p_j,freq_i])[zenith_i]=Sqrt(Mean(Abs((*Jmat_interp[p_i,p_j,freq_i])[zenith_i])^2.))
         
-        horizon_test=where(abs(za_arr) GE 90.,n_horizon_test,complement=pix_use,ncomplement=n_pix)
         horizon_mask=fltarr(psf_image_dim,psf_image_dim)+1
         IF n_horizon_test GT 0 THEN horizon_mask[horizon_test]=0
```
isullivan, line 91: I don't think `horizon_mask` is used anymore, and can be removed.

fhd_core/source_modeling/source_dft_multi.pro
nicholebarry: Sorry for the late reply.

gaussian_x is a vector across all sources, with each element describing the width of the gaussian source sigma in pixels in the x-direction. If the a particular source is a point source, then this is set to 0 for that element. That way, we can easily pick out which sources are gaussians, and filter them into source_dft instead of fast_dft. 

tested and worked.

nicholebarry:
I've taken some time to look over the fast dft, and I don't think I'm going to be able to write in the gaussian source components easily. There's a lot of magic that goes into the fast dft....

So I think I'll make it an issue for someone more willing to look at in the future, if that's okay?

PR #217, opened 2020-06-12
model all 4 source stokes polarizations

### LINKED ISSUE #207: Model all polarizations in 2-pol mode
When `n_pol=2`, FHD only models Stokes I and Q source power. However, there is some contribution to XX and YY visibilities from U and V power. We should model all Stokes polarization modes even when `n_pol` is less than 4.

rlbyrne:
Closes #207

### DIFF
--- a/fhd_core/polarization/stokes_cnv.pro
+++ b/fhd_core/polarization/stokes_cnv.pro
@@ -64,7 +64,6 @@ stokes_inv_term1=[.5,.5,.5,.5]
 stokes_inv_term2=[.5,-.5,-complex(0,.5),complex(0,.5)]
 stokes_list1=[0,0,2,2]
 stokes_list2=[1,1,3,3]
-IF n_pol EQ 1 THEN stokes_list1=(stokes_list2=[0,0,0,0])
 
 IF Keyword_Set(no_dipole_projection_rotation) THEN BEGIN
     ;this is meant as a debugging tool!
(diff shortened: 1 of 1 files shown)


fhd_core/polarization/stokes_cnv.pro
```diff
@@ -124,8 +124,9 @@ IF type EQ 8 THEN BEGIN ;check if a source list structure is supplied
     sx=sx[s_use]
     sy=sy[s_use]
     p_ind=p_ind[s_use]
-    flux_arr=Ptrarr(n_pol,/allocate)
-    flux_pq=Ptrarr(n_pol,/allocate)
+    if keyword_set(inverse) then n_pol_stokes=4 else n_pol_stokes=n_pol
```
isullivan, line 127: You will need a special case for when `n_pol=1`, since on line 67 above `stokes_list` gets set differently then.

rlbyrne: Maybe I'm missing something, but why do we have line 67? I think we should just remove that.

isullivan: Feel free to test the code with a single pol with that line removed. At one time it was necessary to prevent a crash, but I suspect that some of the improvements to the way we treat full polarization since then may have also fixed that.

rlbyrne:
I tested with `n_pol=1` and line 67 is unnecessary. @isullivan, does this look ok now?

Issue #192, opened 2020-02-14
psf_resolution


nicholebarry:
The default is set in multiple locations in the code. In the antenna struct, it's set to 16. In the psf struct, it's set to 1. We should find out which is the true default and delete the other

mwilensky768:
I think this comes down to choosing the more sensible default of the two. If it makes more sense to overresolve by a factor of 16 more often than not overresolving at all, then I say we pick 16. Otherwise, we pick 1.

Based on the fact that it gets set in two places, I think it's possible that one could arrive at either default in any order (one does not necessarily come before the other). Maybe we set them to be identical, instead of deleting one of them?

isullivan:
16 is the real default. The one in defining the psf struct was just a placeholder.

mwilensky768:
I believe this was closed by #219

PR #219, opened 2020-07-21
Change psf_resolution behavior to comply with intended default

mwilensky768:
This is a simple 1-line fix that just makes the defaulting in `fhd_struct_init_psf.pro` match `fhd_struct_init_antenna.pro`. See #192

### DIFF
--- a/fhd_core/beam_modeling/fhd_struct_init_psf.pro
+++ b/fhd_core/beam_modeling/fhd_struct_init_psf.pro
@@ -15,7 +15,7 @@ IF N_Elements(n_pol) EQ 0 THEN n_pol=1 ELSE n_pol=Fix(n_pol)
 IF N_Elements(n_freq) EQ 0 THEN n_freq=1 ELSE n_freq=Fix(n_freq)
 IF N_Elements(freq_cen) EQ 0 THEN freq_cen=Fltarr(n_freq) ELSE freq_cen=Float(freq_cen)
 IF N_Elements(beam_ptr) EQ 0 THEN beam_ptr=Ptr_new() ;will actually have dimensions (npol,nfreq,nbaselines)
-IF N_Elements(psf_resolution) EQ 0 THEN psf_resolution=16. ELSE psf_resolution=Fix(psf_resolution);over-resolution
+psf_resolution=Fix(psf_resolution);over-resolution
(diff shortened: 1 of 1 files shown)


fhd_core/beam_modeling/fhd_struct_init_psf.pro
```diff
@@ -9,13 +9,13 @@ FUNCTION fhd_struct_init_psf,beam_ptr=beam_ptr,complex_flag=complex_flag,$
...
-    psf_image_resolution=psf_image_resolution,psf_image_dim=psf_image_dim   
- 
+    psf_image_resolution=psf_image_resolution,psf_image_dim=psf_image_dim
+
 IF N_Elements(n_pol) EQ 0 THEN n_pol=1 ELSE n_pol=Fix(n_pol)
 IF N_Elements(n_freq) EQ 0 THEN n_freq=1 ELSE n_freq=Fix(n_freq)
 IF N_Elements(freq_cen) EQ 0 THEN freq_cen=Fltarr(n_freq) ELSE freq_cen=Float(freq_cen)
 IF N_Elements(beam_ptr) EQ 0 THEN beam_ptr=Ptr_new() ;will actually have dimensions (npol,nfreq,nbaselines)
-IF N_Elements(psf_resolution) EQ 0 THEN psf_resolution=1 ELSE psf_resolution=Fix(psf_resolution);over-resolution
+IF N_Elements(psf_resolution) EQ 0 THEN psf_resolution=16. ELSE psf_resolution=Fix(psf_resolution);over-resolution
```
isullivan, line 18: I would actually advocate changing this to simply:
`psf_resolution=Fix(psf_resolution);over-resolution`
None of the other 'defaults' in this file are real default values, and a proper cleanup would remove the appearance of defaults and let things crash here if they are not defined previously.

mwilensky768: Do you mean to remove the if/then/else statement entirely, and just have the result of the else?

isullivan: That seems the best to me now, since it will raise the most helpful error message and prevent the code from running with values that might not be what the user wants (typos happen).

PR #227, opened 2020-08-28
beam calculation doubles

nicholebarry:
Add doubles to the impedance, voltage response, and projection. In addition, make the visibility noise calculation double precision. These are important calculations, but small....so there is no noticeable change to runtime or memory required. 

No noticeable change in the power spectra. Perhaps someday this precision will matter.


fhd_core/beam_modeling/general_antenna_response.pro
```diff
@@ -3,16 +3,18 @@ FUNCTION general_antenna_response,obs,antenna,za_arr=za_arr,az_arr=az_arr,psf_im
 n_ant=obs.n_tile
 n_ant_pol=antenna[0].n_pol ;this needs to be the same for all antennas!
 pix_use=*antenna[0].pix_use
-icomp=Complex(0,1)
+icomp=DComplex(0,1)
 c_light_vacuum=299792458.
+Pi=!DPi
```
isullivan, line 8: Very minor point, but I'd suggest just using !DPi throughout, rather than making a new variable.

instrument_config/mwa_dipole_mutual_coupling.pro
```diff
@@ -4,19 +4,17 @@ file_path_Z_matrix=filepath('mwa_ZMatrix.fits',root=rootdir('FHD'),sub='instrume
...
-Zmat_arr=Complexarr(n_ext,n_ant_pol,n_dipole,n_dipole)
-;Zmat_arr_X=Complexarr(n_ext,n_dipole,n_dipole)
-;Zmat_arr_Y=Complexarr(n_ext,n_dipole,n_dipole)
+Zmat_arr=DComplexarr(n_ext,n_ant_pol,n_dipole,n_dipole)
 freq_arr_Zmat=Fltarr(n_ext)
 FOR ext_i=0,n_ext-1 DO BEGIN
     Zmat1=mrdfits(file_path_Z_matrix,ext_i,header,status=status,/silent)
     Zmat=Zmat1[*,*,0]*(Cos(Zmat1[*,*,1])+icomp*Sin(Zmat1[*,*,1]))
-    freq_arr_Zmat[ext_i]=Float(sxpar(header,'FREQ'))
+    freq_arr_Zmat[ext_i]=Double(sxpar(header,'FREQ'))
```
isullivan, line 17: `freq_arr_Zmat` should probably be left as floats, just because that's what the value read from the header will be. If you anticipate that changing at some point in the future, then you should also change its definition on line 12 above to be a double.

isullivan (dismissed):
Looks fine. I had a few minor points for you to consider.

instrument_config/mwa_dipole_mutual_coupling.pro
nicholebarry: The value is read as a double for the frequencies. I've changed the initialisation of the array to be doubles to be more concise.

nicholebarry:
I've also removed some stale, commented out code.

mwilensky768 (approved):
Looks like an equivalent suggestion was followed.

PR #229, opened 2020-09-01
Sim catalogs

mwilensky768:
This PR just makes a new test source that is slightly off of field-center for EoR0 and renames some supposed "RFI" catalogs more appropriately. Since the catalog sources do not have actual RFI-like frequency dependence, but is actually just an ordinary astrophysical point source catalog of simulated point sources, I just renamed them to say what is inside. The power law index for the brightness distribution of the catalogs with "plaw" does resemble that found for RFI sources in [this paper](https://arxiv.org/abs/1610.04696).


nicholebarry (approved):
All good. Please add name changes to inputs.md in the documentation branch. Maybe this week and/or next week we can merge the docu branch

PR #230, opened 2020-09-03
Fragile beam recalculation keyword needs to go

nicholebarry:
The beam recalculation logic is too fragile, and needs to be extracted from the main branch for now.

Some recent requests have indicated that a rethink of some other recalculation keywords would be handy, so we should pick this back up in another branch in the future.


mwilensky768 (approved):
Looks like this just always calculates the beam now, which I am fine with. Reading in a saved beam can be a feature request.

PR #232, opened 2020-09-18
Interpolation test

lmberkhout:


fhd_core/HEALPix/healpix_cnv_generate.pro
```diff
@@ -151,7 +163,9 @@ FOR i=0L,n_img_use-1L DO BEGIN
     *ija[i]=ija0
         
 ENDFOR
-
+;sa=weighting function?
```
isullivan, line 166: `ija` and `sa` are written in the sparse matrix notation of `sprsax2.pro`, in FH/fhd_core/gridding . The `ija`s are the coordinates with data in the sparse matrix, and the `sa`s are the values.

fhd_core/HEALPix/healpix_cnv_generate.pro
```diff
@@ -90,27 +100,29 @@ ENDIF ELSE BEGIN
...
 
 min_bin=Min(Floor(xv_hpx)+dimension*Floor(yv_hpx))>0L
-max_bin=Max(Ceil(xv_hpx)+dimension*Ceil(yv_hpx))<(dimension*elements-1L)
+max_bin=Max(Ceil(xv_hpx)+dimension*Ceil(yv_hpx))<(dimension*elements-1L) 
+; give hist some data, min and max for bins, binsize=1?
 h00=histogram(Floor(xv_hpx)+dimension*Floor(yv_hpx),min=min_bin,max=max_bin,/binsize,reverse_ind=ri00)
 h01=histogram(Floor(xv_hpx)+dimension*Ceil(yv_hpx),min=min_bin,max=max_bin,/binsize,reverse_ind=ri01)
 h10=histogram(Ceil(xv_hpx)+dimension*Floor(yv_hpx),min=min_bin,max=max_bin,/binsize,reverse_ind=ri10)
 h11=histogram(Ceil(xv_hpx)+dimension*Ceil(yv_hpx),min=min_bin,max=max_bin,/binsize,reverse_ind=ri11)
+
```
isullivan, line 123: The above code is calculating the four corners surrounding every pixel, and putting the results in difficult-to-understand lookup tables (the "reverse indices").

isullivan (commented):
I have left some comments that mostly explain what the code is doing at several points. I suspect that you don't want to be using this code at all for what you are doing, though. It has been extensively optimized to make the transformations between ortho-slant and HEALPix fast, but that means that it is inflexible and likely to cause you more trouble than it is worth to get it to do something different.

You would probably have better luck replacing the calls so `healpix_cnv_apply` with a new function that you would write. In the new function, you could take two different obs structures, use `apply_astrometry` to convert the pixel grid of the source obs structure to RA and Dec values, then use `apply_astrometry` again to convert those RA/Dec values to pixel coordinates of the destination obs structure. Then you could probably use one of IDL's interpolation functions (there are two, and they work differently) to convert the values of the source image to values for the destination image.

PR #233, opened 2020-09-23
Allow MWA coarse channel flagging to omit coarse channel edges

isullivan:
Skip flagging the coarse channel edges rather than throwing an error if `channel_edge_flag_width=0` on the command line.


bhazelton (approved):
I tested this and it works.

PR #234, opened 2020-09-25
Fix a bug preventing fitting reflections on multiple cables, improve convergence print statement

bhazelton:
Due to a bug, setting `cal_mode_fit` to an array of multiple cable lengths caused an error. This PR fixes that bug.

It also improves the printed message when a freq, pol doesn't converge in calibration to include the amount the convergence was off by and the convergence tolerance so the user can tell how off they were.


isullivan (approved):
Looks good. Thank you for the fix!

PR #235, opened 2020-09-25
Allow complex and double precision HEALPix

nicholebarry:
This branch has the basic requirements for being able to create cross-pol Healpix cubes. 

It uses the type of the image to determine the type of the Healpix cube (i.e. double, float, real, complex). However, this results in slightly bulky code...checking the type means using a 4 case statement each time. Is there a better way to do it?

This is also not as efficient as it could be. @rlbyrne I remember you saying that the cross-pols are complex conjugates of one another (I think?). Is this always the case, not just for current code, but for most instrumentation and future parameters? If so, we could potentially get away with only writing out three cubes instead of all four.


rlbyrne:
Yes, cross-pols will always be complex conjugates of one another.

I'm concerned about the data volume increase if we make Healpix cubes complex. Does that apply to XX and YY as well? Then those Healpix cubes would be twice the size they need to be. One option is to keep the Healpix cubes real and output XX, YY, the real part of XY, and the imaginary part of XY. This is what we do for the integrated images.

Issue #133, opened 2018-09-10
Model delay filter and edge baselines


nicholebarry:
There is some bad behaviour between baselines which sample inside/outside our designated UV plane as a function of frequency and the model delay filter.

The model delay filter assumes that the whole frequency band is sampled for each non-zero baseline. But that's not the case if a baseline becomes too large to be included at some point as a function of frequency. This hard edge is included, rings, and messes with the delay filter for that baseline as a whole.

An example. Green is a model visibility without the delay filter, black is the same vis with the delay filter.
<img width="605" alt="screen shot 2018-09-10 at 4 35 47 pm" src="https://user-images.githubusercontent.com/5588290/45280584-584bb800-b518-11e8-8fc4-73fbd4453859.png">

We probably need to 1) flag these baselines in both model and data or 2) simulate a larger UV plane for the model, and zero-out accordingly after the delay filter.

nicholebarry:
Tests to investigate:

1) Don't delay filter these visibilities and leave as-is. See what the contamination is like in the EoR window. If it's insignificant, it may be best to just leave these visibilities alone. Include a user message which indicates how many baselines were not delay filtered.
2) Flag these baselines in both model and data. Include a user message which indicates how many extra baselines were flagged, and indicate that a larger UV plane is necessary to include them.

PR #236, opened 2020-09-26
Three rare or minor gridding effects

nicholebarry:
1. Gridding and degridding close to the v=0 line.

Before, whenever a baseline uv-footprint touched the v=0 line, it was flagged. However, I think this is unnecessary. As long as the *center* of the baseline is above v=0, it can be included. This does mean that the footprint will extend into negative v range. In practice, we only grid half of the uv plane and then flip/conjugate and add the two together, which is why I think there was hesitation to include baselines which span both sides of the v=0 line.

I had a good long think about this effect earlier, when I was attempting to do semi-analytic gridding. In that case, a baseline would contribute to *all* of the uv-plane. As long as you grid the top half, but record how that half bleeds downs into the bottom half, the math checks out when you flip/conjugate and add. @miguelfmorales thoughts?

This affect comes in as the `(ymin+(psf_dim/2-1))` lines. I can include PS, but the difference is minor. I don't think there are many baselines which are close to the v=0 line.

2. Double casting the box area pixel total.

When there are many, many baselines, `ii` becomes quite large. `psf_dim3` is a Long. `ii` is a Long. I entered a regime where the multiplication of these two Longs resulted in truncation. By casting psf_dim3 into a Double, I allowed more bits to correct for this issue. This isn't terribly elegant, so I'm open to other suggestions.

3. Flagging baseline extents

Whenever the baseline is too small or too large, it is flagged from the gridding and degridding routines. This was done on a per-frequency basis, which allowed for baselines to disappear halfway through the frequency band. I've changed this so the maximum/minimum extent across the full frequency band is considered when doing this flagging. This allows for a smoother transition along each pixel as a function of frequency, rather than a harsh edge when a baseline disappears. Should fix issue #133.

### DIFF
--- a/fhd_core/gridding/visibility_degrid.pro
+++ b/fhd_core/gridding/visibility_degrid.pro
@@ -102,7 +102,7 @@ dx1dy1_arr = Temporary(dx_arr) * Temporary(dy_arr)
 xmin=Long(Floor(Temporary(xcen))+dimension/2-(psf_dim/2-1))
 ymin=Long(Floor(Temporary(ycen))+elements/2-(psf_dim/2-1))
 
-range_test_x_i=where(((xmin+(psf_dim/2-1)) LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
+range_test_x_i=where(((xmin LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
 range_test_y_i=where(((ymin+(psf_dim/2-1)) LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
 
 IF n_test_x GT 0 THEN xmin[range_test_x_i]=(ymin[range_test_x_i]=-1)
--- a/fhd_core/gridding/visibility_degrid.pro
+++ b/fhd_core/gridding/visibility_degrid.pro
@@ -102,8 +102,8 @@ dx1dy1_arr = Temporary(dx_arr) * Temporary(dy_arr)
 xmin=Long(Floor(Temporary(xcen))+dimension/2-(psf_dim/2-1))
 ymin=Long(Floor(Temporary(ycen))+elements/2-(psf_dim/2-1))
 
-range_test_x_i=where((xmin LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
-range_test_y_i=where((ymin LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
+range_test_x_i=where(((xmin+(psf_dim/2-1)) LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
+range_test_y_i=where(((ymin+(psf_dim/2-1)) LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
 
 IF n_test_x GT 0 THEN xmin[range_test_x_i]=(ymin[range_test_x_i]=-1)
 IF n_test_y GT 0 THEN xmin[range_test_y_i]=(ymin[range_test_y_i]=-1)
@@ -229,10 +229,10 @@ FOR bi=0L,n_bin_use-1 DO BEGIN
           box_matrix, vis_n, beam_int=beam_int, beam2_int=beam2_int, n_grp_use=n_grp_use,degrid_flag=1,_Extra=extra)
     endif else begin
         IF interp_flag THEN $
-          FOR ii=0L,vis_n-1 DO box_matrix[psf_dim3*ii]=$
+          FOR ii=0L,vis_n-1 DO box_matrix[double(psf_dim3)*ii]=$
             interpolate_kernel(*beam_arr[polarization,fbin[ii],baseline_inds[ii]],x_offset=x_off[ii], $
                 y_offset=y_off[ii],dx0dy0=dx0dy0[ii], dx1dy0=dx1dy0[ii], dx0dy1=dx0dy1[ii], dx1dy1=dx1dy1[ii]) $
-        ELSE FOR ii=0L,vis_n-1 DO box_matrix[psf_dim3*ii]=*(*beam_arr[polarization,fbin[ii],baseline_inds[ii]])[x_off[ii],y_off[ii]] ;more efficient array subscript notation
+        ELSE FOR ii=0L,vis_n-1 DO box_matrix[double(psf_dim3)*ii]=*(*beam_arr[polarization,fbin[ii],baseline_inds[ii]])[x_off[ii],y_off[ii]] ;more efficient array subscript notation
     endelse
 
     t4_0=Systime(1)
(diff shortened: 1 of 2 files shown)


miguelfmorales:
Just a meta comment. I think this makes sense as a code change. That said, a baseline kernel crossing to negative uv (through zero) is unphysical, and any uv points within this range really should be tossed from the science interpretation. So I'm wondering about warnings/flags/detailed comments. In particular as the gridding kernel gets big—let's say someone decides to only use the center few degrees of the beam—then the range of short uv baselines that become corrupted by this effect might surprise the user. 

Said another way, this change effectively allows us to avoid a hard filter at small uv and push problems to very small uv. This is good. But the user still needs to know there are weird non-physical effects going on here.

bhazelton:
I'm confused by your comment @miguelfmorales. My understanding was that @nicholebarry was talking about allowing baselines that touch the v=0 line, not the origin (u=0, v=0). I think it is perfectly physical to have baselines that straddle the v=0 line, indeed there can even be baselines that are centered on that line.

miguelfmorales:
We should clear this up. Not gridding across v=0 would just be a bug. That said, it is tricky because you need to treat baselines at +- u differently otherwise you double count. There are a number of _super_ subtle issues, so I want to make sure there is actually a bug. Variants we should make clear:

- flagging a baseline that touches v=0 on either ±u would be a bug.
- conjugating a baseline that touches v=0 at + or - u (but not both) would probably be working as intended
- flagging a baseline that touches the origin is probably an intended original implementation, but changeable.

Said another way, we ant the final uv plane to have nothing special happening at v=0 (either double counting or no counting). This gets tricky at v=0, because it often means including baselines at v=0 on +u and excluding v=0 at -u (or vice versa)—there end up being several ways of implementing this but it interacts with any flipping and summing code _very_ intimately. 

The test is to throw down baselines near v=0 and make sure the final uv plane is correct. A dense uniform grid of baselines not quite aligned with the uv grid (say rotated by a fraction of degree in uv) can be a good test, as the final uv plane should show the correct pattern.

nicholebarry:
I've given this some thought, and I think I have a subtly different line of reasoning and/or I'm not understanding something fully. I'll try to explain.

1) FHD is currently flagging *anything* that touches the v=0 line, even slightly. So something clearly must be done, but the question is what.

2) I'm not fully understanding the difference between +u and -u and how that affects what we grid. 

Let's take the example of one baseline situated somewhere in the uv-plane which has a theoretically infinite kernel (let's say it's two antennas floating in space with no horizon). I would create a full, pixelated uv-plane, and grid down the contribution of the kernel to all pixels. I would then take that grid, flip it, conjugate it, and add it back. I would then have two baselines on my grid, both contributing everywhere, one with a kernel center somewhere above v=0 and one with a kernel center somewhere below v=0. In this example, you would have a small contribution from the global signal into your visibility. That's fine I think...and I'm pretty sure that's the underlying assumption in some global signal experiments which use two closely-spaced antennas. How does the baselines being at +-u have an effect, and how would that be double counting?

fhd_core/gridding/visibility_degrid.pro
```diff
@@ -229,10 +229,10 @@ FOR bi=0L,n_bin_use-1 DO BEGIN
           box_matrix, vis_n, beam_int=beam_int, beam2_int=beam2_int, n_grp_use=n_grp_use,degrid_flag=1,_Extra=extra)
     endif else begin
         IF interp_flag THEN $
-          FOR ii=0L,vis_n-1 DO box_matrix[psf_dim3*ii]=$
+          FOR ii=0L,vis_n-1 DO box_matrix[double(psf_dim3)*ii]=$
```
isullivan, line 232: Instead of casting `psf_dim3` to a double every time it is used, we should change it's type on line 147 where it is defined. Also, could it be a Long64 instead of a double? If that works, it would be preferable.

fhd_core/gridding/visibility_degrid.pro
```diff
@@ -102,8 +102,8 @@ dx1dy1_arr = Temporary(dx_arr) * Temporary(dy_arr)
 xmin=Long(Floor(Temporary(xcen))+dimension/2-(psf_dim/2-1))
 ymin=Long(Floor(Temporary(ycen))+elements/2-(psf_dim/2-1))
 
-range_test_x_i=where((xmin LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
-range_test_y_i=where((ymin LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
+range_test_x_i=where(((xmin LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
+range_test_y_i=where(((ymin+(psf_dim/2-1)) LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
```
isullivan, line 106: I think the existing code is actually OK. It tests whether any pixel of the psf for the given baseline would fall out of the UV plane, and we do want to exclude anything that is outside. Note that xmin and ymin are defined above to be the minimum x and y values of the box of pixels with nonzero values, and are not the coordinates of the baseline. Also note that in this coordinate system baselines with zero length are in the center of the plane, so they should not be affected by this test.

fhd_core/gridding/visibility_grid.pro
```diff
@@ -158,15 +163,15 @@ dx1dy1_arr = Temporary(dx_arr) * Temporary(dy_arr)
 xmin=Long(Floor(Temporary(xcen))+dimension/2-(psf_dim/2-1))
 ymin=Long(Floor(Temporary(ycen))+elements/2-(psf_dim/2-1))
 
-range_test_x_i=where((xmin LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
-range_test_y_i=where((ymin LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
+range_test_x_i=where(((xmin+(psf_dim/2-1)) LE 0) OR ((xmin+psf_dim-1) GE dimension-1),n_test_x)
+range_test_y_i=where(((ymin+(psf_dim/2-1)) LE 0) OR ((ymin+psf_dim-1) GE elements-1),n_test_y)
```
isullivan, line 167: Same comment as in visibility_degrid. I don't think we should change these lines.

isullivan (changes requested):
I'm worried about the changes to the tests for whether baselines are falling off the UV plane. I don't think we should change that code.
Also, if you need to change `psf_dim3` in visibility_degrid, make sure to change it where it is defined instead of casting it to a new type every time it is used. A Long64 would also be preferable to a double. Finally, if you change `psf_dim3` to a Long64 in visibility_degrid, please also change it in visibility_grid.

nicholebarry:
Thanks for your patience, all. I understand now why my thinking was incorrect. Rare (de)grid effect 1 was a misunderstanding.

isullivan (approved):
This now looks good to me.

bhazelton:
I recently found that a diff power spectrum for some minor calibration speedups that Ian is working on had much more difference than I had expected. I traced it back to the changes in this PR (I was inadvertently comparing runs from before and after this PR was merged). I made a diff plot that just has the differences from this PR so I could compare and am attaching it here for future reference.

fhd_van_vleck_noimgclip_dft_averemove_swbh_dencorr__grid2_beforegriddingpr_1061316296_minus_grid2_1061316296_2dkpower

I think the changes in this PR are correct, but they do have a significant affect on the power spectrum. They seem to increase the power in the wedge for the model and in the wedge and just above the wedge in the dirty and residual by 1e6 to 1e9.

PR #237, opened 2020-10-02
Rearrange (de)grid and remove excess time calls for speed up

nicholebarry:
I've made some slight modifications to gridding and degridding for a modest speed-up of about 10 minutes total on a typical run. 

Changes include:

- Reducing `interp` IF statements by grouping them together. I only have had to do a small amount of code redundancy to make this happen
- Removing internal time calls. Apparently this was a time sink
- Removing the FLOOR call in an index calculation. It was unnecessary as far as I can tell.
- General aesthetic cleanup

Running on this branch and the master branch produced the same exact power spectra (tested with the `interp` keyword)

### DIFF
--- a/fhd_core/gridding/visibility_degrid.pro
+++ b/fhd_core/gridding/visibility_degrid.pro
@@ -137,18 +137,9 @@ CASE 1 OF
 ENDCASE
 arr_type=Size(init_arr,/type)
 
-time_check_interval=Ceil(n_bin_use/10.)
-t1=0
-t2=0
-t3=0
-t4=0
-t5=0
 image_uv_use=image_uv
 psf_dim3=psf_dim*psf_dim
 
-;pdim=size(psf_base,/dimension)
-;psf_base_dag=Ptrarr(pdim,/allocate)
-;FOR pdim_i=0L,Product(pdim)-1 DO *psf_base_dag[pdim_i]=Conj(*psf_base[pdim_i])
 IF Keyword_Set(n_spectral) THEN BEGIN
     prefactor=Ptrarr(n_spectral)
     FOR s_i=0,n_spectral-1 DO prefactor[s_i]=Ptr_new(deriv_coefficients(s_i+1,/divide_factorial))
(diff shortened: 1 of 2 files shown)


fhd_core/gridding/visibility_degrid.pro
```diff
@@ -174,69 +160,74 @@ FOR bi=0L,n_bin_use-1 DO BEGIN
...
+            ENDELSE
         
-        vis_n=Long64(n_xyf_bin)
-    ENDIF ELSE BEGIN
-        ind_remap_flag=0
-        bt_index = inds / n_freq_use
+            vis_n=Long64(n_xyf_bin)
+        ENDIF ELSE BEGIN
+            ind_remap_flag=0
+            bt_index = inds / n_freq_use
```
isullivan, line 215: Use `//` for integer division

fhd_core/gridding/visibility_grid.pro
```diff
@@ -282,96 +265,101 @@ FOR bi=0L,n_bin_use-1 DO BEGIN
...
+            IF model_flag THEN FOR rep_ii=0,n_rep-1 DO $
+                model_box[repeat_i[rep_ii]]=Total(model_box1[xyf_ui0[rep_ii]:xyf_ui[rep_ii]])
+
+            vis_n=n_xyf_bin
+        ENDIF ELSE BEGIN
+            rep_flag=0
+            IF model_flag THEN model_box=model_use[inds]
+            vis_box=vis_arr_use[inds]
+            psf_weight=Replicate(1.,vis_n)
+            bt_index = inds / n_freq_use
```
isullivan, line 336: Use `//` for integer division.

isullivan (approved):
These changes look fine, though please do make sure the baseline indices calculations strictly use integer division (I think I commented on all of the instances, but it wouldn't hurt to double-check).

nicholebarry:
@isullivan I'm not sure the `//` is IDL, and I couldn't find another substitute.

isullivan:
> @isullivan I'm not sure the `//` is IDL, and I couldn't find another substitute.

You are completely correct! As long as both inputs are integers in IDL, the output will be an integer.

PR #238, opened 2020-10-04
Speed up Healpix export code 

nicholebarry:
Fairly straightforward speed-up of Healpix export code. There was an unnecessary middle-man pointer when reorganizing the Healpix cubes into vectors. Since the Healpix cubes are so large, removing this excess reduced the runtime by about 6 minutes and reduced memory needed by a few Gb. A slight reordering of a `for` loop was also done in the even/odd split, but this probably is a minuscule effect. 

There is no change in the power spectra

### DIFF
--- a/fhd_core/HEALPix/healpix_snapshot_cube_generate.pro
+++ b/fhd_core/HEALPix/healpix_snapshot_cube_generate.pro
@@ -143,86 +143,62 @@ PRO healpix_snapshot_cube_generate,obs_in,status_str,psf_in,cal,params,vis_arr,v
     t_split+=t_split1
     IF dirty_flag THEN BEGIN
       dirty_arr1=residual_arr1
-;      residual_arr1=Ptrarr(size(residual_arr1,/dimension),/allocate)
       residual_flag=0
     ENDIF ELSE residual_flag=1
     nf_vis=obs.nf_vis
     nf_vis_use=Lonarr(n_freq_use)
     FOR freq_i=0L,n_freq_use-1 DO nf_vis_use[freq_i]=Total(nf_vis[freq_i*n_avg:(freq_i+1)*n_avg-1])
     
-    IF residual_flag THEN residual_hpx_arr=Ptrarr(n_pol,n_freq_use,/allocate)
-    model_hpx_arr=Ptrarr(n_pol,n_freq_use,/allocate)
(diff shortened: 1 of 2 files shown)


mwilensky768 (dismissed):
A bit dizzying to look at, but it looks like all lines are accounted for, and the testing seems solid. Any idea what the middle-man pointers were there for? I'd have to imagine they were intentional.

fhd_core/HEALPix/healpix_snapshot_cube_generate.pro
```diff
@@ -143,86 +143,62 @@ PRO healpix_snapshot_cube_generate,obs_in,status_str,psf_in,cal,params,vis_arr,v
...
-          dirty_cube=fltarr(n_hpx,n_freq_use)
-            ;write index in much more efficient memory access order
-          FOR fi=Long64(0),n_freq_use-1 DO dirty_cube[n_hpx*fi]=Temporary(*dirty_hpx_arr[pol_i,fi])
-        ENDIF
         
-        IF model_flag THEN BEGIN
-          model_cube=fltarr(n_hpx,n_freq_use)
-          FOR fi=Long64(0),n_freq_use-1 DO model_cube[n_hpx*fi]=Temporary(*model_hpx_arr[pol_i,fi])
+        IF Keyword_Set(save_imagecube) THEN BEGIN
+            save, filename = imagecube_filepath[iter], dirty_arr1, residual_arr1, model_arr1, weights_arr1, variance_arr1, beam_arr, nf_vis_use, obs_out, /compress
```
isullivan, line 182: This needs to get moved outside of the polarization `for` loop, or else the same files will be written multiple times.

isullivan (dismissed):
The pointers were there so that the healpix cubes could be calculated for all polarizations, then reformatted in a faster indexing order and saved. The reduced memory usage comes from only calculating one polarization at a time, and reformatting and saving it, before moving on to the next polarization.

fhd_core/HEALPix/healpix_snapshot_cube_generate.pro
nicholebarry: Good catch

mwilensky768 (approved):
Thanks for checking in @isullivan Looks like things are living in the right loops now.

PR #240, opened 2020-10-31
one-mean to zero-mean for regular cable fitting

nicholebarry:
When incorporating Wenyang's auto ratio branch, I needed to change the convention used in removing the polyfit contribution before cable fitting. This didn't properly percolate to the regular cable fitting.

Therefore, we were taking a hyperfine DFT over something that had a mean of 1 instead of a mean of 0. The associated ringing was rendering the cable reflections useless. 

One of those cases where essentially a `gain = gain - 1` changes the PS significantly....


nicholebarry:
See pull request #245

PR #241, opened 2020-10-31
Add simple function to define named structures

isullivan:
This simple function creates a placeholder structure of the specified type, and deletes it. This defines the structure parameters for the current IDL session and ensures consistency when restoring structures of that type.


bhazelton (approved):
This fixed the bug I was seeing. Thanks!

Issue #243, opened 2020-12-10
Possible Faulty Logic in source_dft_multi


mwilensky768:
This line is being executed in `source_dft_multi` when I have `dft_threshold=1` and `gaussian_source_models` on default setting (should be off).

https://github.com/EoRImaging/FHD/blob/c9d656250a19fe9ca2c1f6f027eee467cbcc0370/fhd_core/source_modeling/source_dft_multi.pro#L213

I think this line implicitly assumes `gaussian_source_models` has been set, since it takes elements of `gaussian_x` and `gaussian_y`. In my version, it is erroring, because these variables are never assigned. This is probably because `gaussian_source_models` is indeed off, and so these variables are never set. 

Should this line only be executed if `gaussian_source_models` is set? In either case, some additional logic is required.

@nicholebarry

nicholebarry:
Solved with #245

PR #244, opened 2020-12-10
Fix bug for missing gauss param for edge sources

nicholebarry:
There was some faulty logic as detailed in Issue #243 . I fixed it almost a month ago, but never set up a pull request. Apologies.

@mwilensky768 see if this works for you.


mwilensky768:
Testing this via #245 since I am also using those changes.

PR #245, opened 2020-12-10
Cal fix and speedup

isullivan:
Combining commits from three branches. From @bhazelton: One of them addresses Mike’s issue above (edge_source_fix), one fixes a cable reflection fitting bug (cable_reflection_fix) and one implements Ian’s new calibration speedup (calibration_speedup).

### DIFF
--- a/fhd_core/calibration/fhd_struct_init_cal.pro
+++ b/fhd_core/calibration/fhd_struct_init_cal.pro
@@ -33,7 +33,6 @@ IF N_Elements(max_cal_baseline) EQ 0 THEN max_cal_baseline=obs.max_baseline ELSE
 IF N_Elements(cal_time_average) EQ 0 THEN cal_time_average=1 ;time average visibilities before calculating calibration solutions by default
 IF N_Elements(min_cal_solutions) EQ 0 THEN min_cal_solutions=5
 IF N_Elements(max_cal_iter) EQ 0 THEN max_cal_iter=100L
-IF N_Elements(phase_fit_iter) EQ 0 THEN phase_fit_iter=Long((Floor(max_cal_iter/4.)<Floor(Sqrt(max_cal_iter)))>4) ELSE phase_fit_iter=Long(phase_fit_iter)
 IF N_Elements(ref_antenna) EQ 0 THEN ref_antenna=1L
 ref_antenna_name=(*obs.baseline_info).tile_names[ref_antenna]
 IF N_Elements(cal_convergence_threshold) EQ 0 THEN cal_convergence_threshold=1E-7
@@ -59,6 +58,7 @@ IF N_Elements(cal_mode_fit) EQ 0 THEN cal_mode_fit=0.
 IF N_Elements(use_adaptive_calibration_gain) EQ 0 THEN use_adaptive_calibration_gain=0
 ;The relative weight to give the old calibration solution when averaging with the new. 
 IF N_Elements(calibration_base_gain) EQ 0 THEN calibration_base_gain=1.
+IF N_Elements(phase_fit_iter) EQ 0 THEN phase_fit_iter=Long((calibration_base_gain*4)>4) ELSE phase_fit_iter=Long(phase_fit_iter)
 convergence=Ptrarr(2)
 mode_params=Ptrarr(n_pol,n_tile)
 auto_params=Ptrarr(2)
--- a/fhd_core/calibration/fhd_struct_init_cal.pro
+++ b/fhd_core/calibration/fhd_struct_init_cal.pro
@@ -8,7 +8,7 @@ FUNCTION fhd_struct_init_cal, obs, params, skymodel, gain_arr_ptr=gain_arr_ptr, 
     calibration_polyfit=calibration_polyfit, bandpass_calibrate=bandpass_calibrate, $
     cal_mode_fit=cal_mode_fit, file_path_fhd=file_path_fhd, transfer_calibration=transfer_calibration, $
     calibration_auto_initialize=calibration_auto_initialize, cal_gain_init=cal_gain_init, $
-    phase_fit_iter=phase_fit_iter, cal_amp_degree_fit=cal_amp_degree_fit,cal_phase_degree_fit=cal_phase_degree_fit,$
+    phase_fit_iter=phase_fit_iter, cal_amp_degree_fit=cal_amp_degree_fit,cal_phase_degree_fit=cal_phase_degree_fit, $
     use_adaptive_calibration_gain=use_adaptive_calibration_gain, calibration_base_gain=calibration_base_gain,_Extra=extra
 
 IF N_Elements(obs) EQ 0 THEN fhd_save_io,0,obs,var='obs',/restore,file_path_fhd=file_path_fhd
@@ -65,19 +65,19 @@ mode_params=Ptrarr(n_pol,n_tile)
 auto_params=Ptrarr(2)
 auto_scale=Fltarr(2)
 
-cal_struct={n_pol:n_pol, n_freq:n_freq, n_tile:n_tile, n_time:n_time, uu:u_loc, vv:v_loc,$
-    auto_initialize:auto_initialize, max_iter:max_cal_iter, phase_iter:phase_fit_iter,$
-    tile_A:tile_A, tile_B:tile_B, tile_names:tile_names, bin_offset:bin_offset, freq:freq, gain:gain_arr_ptr,$
-    adaptive_gain:use_adaptive_calibration_gain,base_gain:calibration_base_gain,$
-    gain_residual:gain_residual, auto_scale:auto_scale, auto_params:auto_params, cross_phase:0.0, stokes_mix_phase:0.0,$
-    min_cal_baseline:min_cal_baseline, max_cal_baseline:max_cal_baseline, n_vis_cal:n_vis_cal,$
-    time_avg:cal_time_average, min_solns:min_cal_solutions, ref_antenna:ref_antenna,$
-    polyfit:calibration_polyfit, amp_degree:cal_amp_degree_fit, phase_degree:cal_phase_degree_fit,$
-    amp_params:amp_params, phase_params:phase_params,$
-    mean_gain:Fltarr(n_pol), mean_gain_residual:Fltarr(n_pol), mean_gain_restrict:Fltarr(n_pol),$
-    stddev_gain_residual:Fltarr(n_pol), bandpass:bandpass_calibrate, mode_fit:cal_mode_fit,$
+cal_struct={n_pol:n_pol, n_freq:n_freq, n_tile:n_tile, n_time:n_time, uu:u_loc, vv:v_loc, $
+    auto_initialize:auto_initialize, max_iter:max_cal_iter, phase_iter:phase_fit_iter, $
+    tile_A:tile_A, tile_B:tile_B, tile_names:tile_names, bin_offset:bin_offset, freq:freq, gain:gain_arr_ptr, $
+    adaptive_gain:use_adaptive_calibration_gain, base_gain:calibration_base_gain, $
+    gain_residual:gain_residual, auto_scale:auto_scale, auto_params:auto_params, cross_phase:0.0, stokes_mix_phase:0.0, $
+    min_cal_baseline:min_cal_baseline, max_cal_baseline:max_cal_baseline, n_vis_cal:n_vis_cal, $
+    time_avg:cal_time_average, min_solns:min_cal_solutions, ref_antenna:ref_antenna, $
     ref_antenna_name:ref_antenna_name, conv_thresh:cal_convergence_threshold, $
     convergence:convergence, n_converged:Lonarr(n_pol)-1, conv_iter:conv_iter, $
+    polyfit:calibration_polyfit, amp_degree:cal_amp_degree_fit, phase_degree:cal_phase_degree_fit, $
+    amp_params:amp_params, phase_params:phase_params, $
+    mean_gain:Fltarr(n_pol), mean_gain_residual:Fltarr(n_pol), mean_gain_restrict:Fltarr(n_pol), $
+    stddev_gain_residual:Fltarr(n_pol), bandpass:bandpass_calibrate, mode_fit:cal_mode_fit, $
     mode_params:mode_params, cal_origin:calibration_origin, skymodel:skymodel}
 RETURN,cal_struct
 END
--- a/fhd_core/calibration/fhd_struct_init_cal.pro
+++ b/fhd_core/calibration/fhd_struct_init_cal.pro
@@ -60,6 +60,7 @@ IF N_Elements(use_adaptive_calibration_gain) EQ 0 THEN use_adaptive_calibration_
 IF N_Elements(calibration_base_gain) EQ 0 THEN calibration_base_gain=1.
 IF N_Elements(phase_fit_iter) EQ 0 THEN phase_fit_iter=Long((calibration_base_gain*4)>4) ELSE phase_fit_iter=Long(phase_fit_iter)
 convergence=Ptrarr(2)
+conv_iter=Ptrarr(2)
 mode_params=Ptrarr(n_pol,n_tile)
 auto_params=Ptrarr(2)
 auto_scale=Fltarr(2)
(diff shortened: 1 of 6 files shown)


fhd_core/source_modeling/source_dft_multi.pro
```diff
@@ -129,13 +136,8 @@ IF Keyword_Set(n_spectral) THEN BEGIN
...
         model_uv_arr=Ptrarr(n_pol,n_spectral+1)
-        if ~keyword_set(gaussian_source_models) then begin
-            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix),$
-                n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
-        endif else begin
-            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix) OR $
-                gaussian_x,n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
-        endelse
+        edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix) OR $
+            gaussian_x,n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
```
isullivan, line 140: I'm sure I'm missing something here, but why does this only tests `gaussian_x` and not also `gaussian_y`?

fhd_core/source_modeling/source_dft_multi.pro
```diff
@@ -201,13 +203,8 @@ ENDIF ELSE BEGIN
...
         model_uv_arr=Ptrarr(n_pol)
-        if ~keyword_set(gaussian_source_models) then begin
-            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix),$
-                n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
-        endif else begin
-            edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix) OR $
-                gaussian_x,n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
-        endelse
+        edge_i = where((x_vec LT dft_edge_pix) OR (y_vec LT dft_edge_pix) OR (dimension-x_vec LT dft_edge_pix) OR (elements-y_vec LT dft_edge_pix) OR $
+            gaussian_x,n_edge_pix, complement=center_pix_i, ncomplement=n_center_pix)
```
isullivan, line 207: As above, why no test for `gaussian_y`?

fhd_core/calibration/vis_cal_polyfit.pro
```diff
@@ -274,7 +274,8 @@ FUNCTION vis_cal_polyfit,cal,obs,cal_step_fit=cal_step_fit,cal_neighbor_freq_fla
             gain_temp=rebin(transpose(reform(resautos[freq_use])),nmodes,nf_use)
 
           ENDIF ELSE BEGIN
-            gain_temp=rebin_complex(transpose(reform(gain_arr[freq_use,tile_i])),nmodes,nf_use) ; dimension manipulation, add dim for mode fitting
+            ; dimension manipulation, add dim for mode fitting
+            gain_temp=rebin_complex(transpose(reform(gain_arr[freq_use,tile_i] - mean(gain_arr[freq_use,tile_i]))),nmodes,nf_use)
```
isullivan, line 278: Please add a code comment explaining why the mean is being subtracted.

isullivan (commented):
Someone else should look through my calibration speedup changes.

fhd_core/source_modeling/source_dft_multi.pro
nicholebarry: I'm only testing for the shape in the x direction because it would be unphysical to have an infinitely flat cigar-shaped source. I can add the gaussian_y logic if you think it's worth it (it's just bulky, that's really the only reason against)

isullivan: OK. Please add a code comment explaining that choice.

nicholebarry (commented):
Code looks sound.

Things to investigate and/or discuss:
1) Tolerance levels. Can there be a convergence "noise" whilst the trend is gently converging? It appears as though some divergence values have been chosen to account for this. Are these values good for non-MWA instruments too?
2) In the unlikely event that a gain is a NAN, we should set it to zero (I assume) somewhere.
3) Testing on good obs and bad obs to see if there is a bias.

mwilensky768:
Here is the difference power spectrum for `master` at `a2dec7` and the PR at `4c5f58`.

_noimgclip_dft_averemove_swbh_dencorr__master_1094485088_minus_PR_1094485088_2dkpower

nicholebarry:
Does this include the "Revert to orig sign convention in exp and match" commit from 3 days ago?

mwilensky768:
Yes.

nicholebarry:
Okay, cool. So I assume you ran with some form of pre-Wenyang calibration, in which case the cable reflection was better accounted for in this PR. Would it be possible for you to run with `auto-ratio` calibration for both as well (just to remove the contribution of the cable reflection)?

Also, can you post cal structures and/or plots on the differences of the convergence levels and overall "raw" gains (i.e. per-freq gains directly from the subroutine prior to fitting). Also, what was the time/mem results for each test?

mwilensky768:
Yes, pre-Wenyang cal. Will get to work on the other plots/metrics.

mwilensky768:
The timing/memory for 10 cores of an r.8xlarge instance on AWS is as follows

master: 63G RAM, 223 min
PR: 63G RAM, 180 min

I think there is an underflow issue cropping up with the cal convergences which is making it seem like 3/4 of them are identically 0. It does not match the flags read in by a UVCal object, and the cal plots for the obs look too sane for on average 3/4 of the band per antenna being flagged. Will post plots representing this thinking on request.

nicholebarry:
I've tested this is in both simulation and data. They both tell a very similar story, so I'll show just simulation for simplicity. 

Simulation specifics: Input visibilities have ~50,000 sources and a smooth beam, and the model has far less sources and the fft of the instrumental beam. The simulation is allowed to fit a calibration amp and phase per frequency (the classic overfitting scenario).

There are three types of tests here:
1. Original linear least squares fitting
2. This updated branch, but still using the original linear least squares fitting.
3. This updated branch with the new adaptive gain fitting technique.

In order to test the original least squares fitting (1), I've just reverted the subroutine. Thus, I really am only comparing the fitting technique. The updated branch with original linear least squares fitting (2) is not quite the same as (1) because there is a smarter calculation of the threshold criteria.

First, a test on how many iterations till the threshold criteria is reached for all tiles and for one frequency channel. Black lines are each tile's convergence in that iteration. Green is the mean over all tiles. The first 4 iterations fit for just the phase, hence the discontinuity.
1. Original linear least squares fitting
old_conv
2. This updated branch, but still using the original linear least squares fitting.
new_conv
3. This updated branch with the new adaptive gain fitting technique.
adaptive_conv
Takeaways: The smarter calculation of the threshold criteria (2) is better than user-input threshold criteria (1). Oddly enough, the adaptive gain (3) converged slower.

Second, a test on what the converged result was for both amp and phase. The changes are too small to see by eye, so I am comparing the differences. Black is each tile's difference, and green is the standard deviation over all tiles.
* Amplitude: This updated branch, but still using the original linear least squares fitting (2) MINUS Original linear least squares fitting (1)
new_minus_old_amp_sim
* Amplitude: This updated branch with the new adaptive gain fitting technique (3) MINUS This updated branch, but still using the original linear least squares fitting (2)
adaptive_minus_new_amp_sim
* Phase: This updated branch, but still using the original linear least squares fitting (2) MINUS Original linear least squares fitting (1)
new_minus_old_phase_sim
* Phase: This updated branch with the new adaptive gain fitting technique (3) MINUS This updated branch, but still using the original linear least squares fitting (2)
adaptive_minus_new_phase_sim
Takeaways: The differences are close to floating point, but there seem to be a few outliers, particularly when comparing against the adaptive gain technique (3).

Third, the power spectra difference.
* Original linear least squares fitting (1) MINUS This updated branch, but still using the original linear least squares fitting (2) 
fhd_nb_sim_regbeam_noEoR_overcal_noimgclip_dft_averemove_swbh_dencorr__oldsubcheck_1061316296_minus_newsubcheck_1061316296_2dkpower
*  This updated branch, but still using the original linear least squares fitting (2) MINUS This updated branch with the new adaptive gain fitting technique (3)
fhd_nb_sim_regbeam_noEoR_overcal_noimgclip_dft_averemove_swbh_dencorr__newsubcheck_1061316296_minus_newsubcheck_adaptive_1061316296_2dkpower
Takeaways: the power spectrum difference is very small (EoR level) and seems to get better with each test in the window: (2) is better than (1), (3) is better than (2). It's unclear to me why it is better, perhaps @isullivan will have a theory.

One final note: 
The convergence vs iteration test I showed above is just for one frequency. In fact, not all frequencies converge. The number of converged frequencies is different between each test. Again, this is simulation with no flagged frequencies.
1. Original linear least squares fitting: 0 (xx) and 0 (yy) because I set the convergence threshold too low.
2. This updated branch, but still using the original linear least squares fitting: 384 (xx) and 383 (yy)
3. This updated branch with the new adaptive gain fitting technique: 378 (xx) and 377 (yy)

Therefore, it seems as though the new convergence threshold calculation in (2) is a win, and that the adaptive gain fitting technique (3) is promising but may need tweaking. I will look into the failed frequencies to see if I can't pinpoint why.

bhazelton:
This is fascinating! Thanks for looking into it so thoroughly @nicholebarry!

isullivan:
This is a great analysis @nicholebarry! A few comments: 
The initial blip around iteration 5 is when we switch from only fitting the phase to the full gain calculation.
I'm very happy to see how the new threshold is working, that convergence plot looks great!
Since these are simulations, I assume the large differences in the final amplitude and phase plots for (3) around 185 and 195 MHz are unphysical, right? I am confused about the power spectrum improvement in this case. Are these the frequencies that failed to converge, maybe? I suspect the oscillations in the convergence plot are prematurely triggering the end of calibration, which suggests that the maximum gain is allowed to be too aggressive. I also suspect that these oscillations may be the reason (3) has fewer converging frequencies, since there may be a jump up that exits calibration just above the threshold used to indicate a converging or diverging solution.

nicholebarry:
Three issues:

1. The base gain is defaulted to 1 in the calibration init procedure. The adaptive gain procedure then checks to see if base gain is set, and if not, it is set to 0.5. This defaulting in the adaptive gain procedure will never happen due to the prior default in the init. Rerunning tests with the base gain set to 0.5 increases speed, but also increases the number of diverged frequency channels. Here is an example frequency convergence using a base gain of 0.5.
adaptive_conv_half
And of an example of frequency convergence using a base gain of 0.75 (in other words, slower but steadier convergence). For this frequency, it actually converged *faster* than the base gain of 0.5, but if you look closely, it has a slightly flatter slope.
adaptive_75_update_conv

2. Convergence checking begins where `IF i GT phase_fit_iter THEN BEGIN` where `i` is the iteration and `phase_fit_iter` is the number of iterations of fitting just the phase. I suggest changing this to `IF i GT phase_fit_iter+divergence_history THEN BEGIN` where `divergence_history` is number of prior iterations to compare to in order to see if the current iteration is diverging. Since there is an expected discontinuity in the convergence once amplitudes are included, some of the divergence calculations are being skewed. This solves approx 2 to 4 diverged frequency channels which failed converging at 1e-2 when it should have reached 1e-7.

3. Adding an additional logic condition reduces a lot of the "diverged channels". In reality, hitting a threshold of 1e-7 (default) is extremely hard, as showcased is the original convergence plot (first plot above) which never truly converged. We are converging (for the most part) with the update because there is a hard condition (max deviation) and a soft condition (mean deviation) and if we start to diverge, we check to see if the soft condition has been met and call it good. However, we don't check to see if the previous iteration, the one before the start of divergence, met the soft condition. I've added logic to check that and back up the iteration if it has. 

Incorporating these changes, here are the number of converged channels and mean iterations till convergence for:
* 0.5 base gain: 381 (xx), 377 (yy), 32.6 mean iterations till convergence 
* 0.75 base gain: 384 (xx) 384 (yy), 33.6 mean iterations till convergence

Given this, I'm pushing these changes in addition to defaulting the base gain to 0.75.

fhd_core/calibration/fhd_struct_init_cal.pro
```diff
@@ -54,22 +54,30 @@ phase_params=Ptrarr(n_pol,n_tile)
 
 IF N_Elements(bandpass_calibrate) EQ 0 THEN bandpass_calibrate=1
 IF N_Elements(cal_mode_fit) EQ 0 THEN cal_mode_fit=0.
+;whether to use a Kalman Filter to adjust the gain to use for each iteration of calculating calibration
+IF N_Elements(use_adaptive_calibration_gain) EQ 0 THEN use_adaptive_calibration_gain=0
+;The relative weight to give the old calibration solution when averaging with the new. 
+IF N_Elements(calibration_base_gain) EQ 0 THEN calibration_base_gain=1.
+IF N_Elements(phase_fit_iter) EQ 0 THEN phase_fit_iter=Long((calibration_base_gain*4)>4) ELSE phase_fit_iter=Long(phase_fit_iter)
```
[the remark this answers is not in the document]
isullivan: It is set to ensure there are at least four iterations of phase-only calibration before we start fitting amplitude, or more if the gain is high (which makes convergence slower.

Coming back to this, I am actually not at all sure we want *any* iterations of phase-only calibration. It was part of the best practices I picked up from calibration experts in 2012, but I am not convinced any more that it is a good thing. Could you do one of your tests with `phase_fit_iter` hard-coded to 0?

fhd_core/calibration/vis_calibrate_subroutine.pro
```diff
@@ -180,27 +190,69 @@ FUNCTION vis_calibrate_subroutine,vis_ptr,vis_model_ptr,vis_weight_ptr,obs,cal,p
...
             endif
-            conv_test[fii,i]=Max(Abs(gain_curr-gain_old)*weight_invert(Abs(gain_old)))
-            IF i GT phase_fit_iter THEN IF conv_test[fii,i] LE conv_thresh THEN BEGIN
-                n_converged += 1
-                BREAK
+            convergence_strict = Max(Abs(gain_curr-gain_old)*weight_invert(Abs(gain_old)))
+            convergence_loose = Mean(Abs(gain_curr-gain_old)*weight_invert(Abs(gain_old)))
+            convergence_list[i] = convergence_strict
+            conv_test[fii,i] = convergence_strict
+            IF i GT phase_fit_iter+divergence_history THEN BEGIN
```
isullivan, line 208: We should add another test here. If the solution diverges above `sqrt(conv_thresh)` then we should keep going, but save the last solution before that point. If the next solution from the next iteration is not better then we exit and use our saved solution, but if it does improve again we can keep going.

isullivan (commented):
I added a couple more ideas based on the calibration plots you shared. The first asks the question of whether we actually need phase-only solutions at first, and the second tries to fix the solutions you saw that diverged early. I can write the code for the new divergence code if you like.
Everything here looks fine, and I could also do the new code on a new branch, since this is well tested. I would mark this pull request as 'approved', but can't since I was the author.

fhd_core/calibration/fhd_struct_init_cal.pro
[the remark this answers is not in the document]
nicholebarry: For the MWA EoR data, yes, it's totally unnecessary. The phase-first iterations are for finding the local minima (the actual trough in which to pursue solutions, I don't know what that's called...) in telescopes with much less stable phase solutions, and is still part of best-practices for multi-purpose, multi-instrument low-frequency analysis. I'd say keep it in, and maybe play around with eor wrapper defaults in order to get a slight speed-up. Will do a test with it set to 0 to see how much speed up we get

nicholebarry (approved):

Issue #242, opened 2020-11-18
Minimum baseline calculation in obs init vs degridding/gridding


nicholebarry:
In calculating the minimum baseline length in wavelengths, we use the u and v speed of light travel time and the frequency of the measurement. We also cast the calculation of the baseline length array into a Float (probably because the precision isn't needed and it's a very large array, but I could be wrong).

However, * where * the casting happens in the calculation is different for gridding/degridding and the initialisation of the obs structure, so the answers are slightly different. For the MWA, the minimum baseline length in lambda is calculated to be 4.30430 in the obs structure initialisation and 4.3042951 in gridding/degridding. This is a problem because the obs.minimum_baseline is used in logic in gridding/degridding to flag individual visibilities which are below the threshold.

This small discrepancy results in different flagging between even and odd eppsilon inputs.

isullivan:
Good catch! We should make the calculation for the value in the obs structure match what is used for gridding.

nicholebarry:
Okay, will do. I may also try to see where full precision would be worthwhile (for example, if double_precision keyword is set)

isullivan:
Sorry, one more comment: While I think the calculation in the obs structure should match what's used in gridding/degridding, those should then just use the value stored in the obs structure so that we are guaranteed to use the same value everywhere.

PR #248, opened 2020-12-19
Min/max baseline comparisons now double

nicholebarry:
Previously, we would cast our baseline calculations to floating point. However, where the cast happened would change the result. This showed up as a difference in the number of visibilities going into the even/odd cubes due to an inconsistent minimum baseline (see #242).

I've modified the code so that the baseline calculations (i.e. calculating the min/max baseline in the obs structure, calculating the baselines in gridding/degridding) so that they are done natively at double precision. I did not notice a change in mem/time requirements, but it should be something to be paid attention to for extreme baseline numbers (perhaps for the SKA?). The number of visibilities are now the same between even/odd cubes.

Recent gridding pushes were already at double precision.

### DIFF
--- a/fhd_core/setup_metadata/fhd_struct_init_obs.pro
+++ b/fhd_core/setup_metadata/fhd_struct_init_obs.pro
@@ -190,7 +190,7 @@ struct={code_version:String(code_version),instrument:String(instrument),obsname:
     phasera:meta.phasera,phasedec:meta.phasedec,orig_phasera:meta.orig_phasera,orig_phasedec:meta.orig_phasedec,$
     n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:n_time_cut,$
     n_vis:Long(n_vis),n_vis_in:Long(n_vis_in),n_vis_raw:Long(n_vis_raw),nf_vis:Long(n_vis_arr),primary_beam_area:Ptrarr(4),primary_beam_sq_area:Ptrarr(4),pol_names:pol_names,$
-    jd0:meta.jd0,max_baseline:max_baseline,min_baseline:min_baseline,delays:meta.delays,lon:meta.lon,lat:meta.lat,alt:meta.alt,$
+    jd0:meta.jd0,max_baseline:Double(max_baseline),min_baseline:Double(min_baseline),delays:meta.delays,lon:meta.lon,lat:meta.lat,alt:meta.alt,$
     freq_center:Float(freq_center),freq_res:Float(freq_res),time_res:Float(meta.time_res),astr:meta.astr,alpha:Float(spectral_index),$
     residual:0,vis_noise:noise_arr,baseline_info:Ptr_new(arr),meta_data:meta_data,meta_hdr:meta_hdr,$
     degrid_spectral_terms:degrid_spectral_terms,grid_spectral_terms:grid_spectral_terms,grid_info:grid_info,healpix:healpix}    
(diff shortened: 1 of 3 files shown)


nicholebarry:
Tested many times over, and I get a nice agreement on the number of visibilities in even/odd cubes for multiple pointings/obs/etc. 

I battled with one small error for quite some time, which turned out to be a casting bug. The value in `vis_n_full` was a long integer, and so dividing it by a float would result in losing a baseline or two at the end of the list. This only was used in high memory situations in the degridding, so a few baselines at approx less than 10 wavelengths were affected. 

This is the result before the bug fix:
fhd_nb_model_gaussdecomp_update_nofreqdep_corrmin_double_uvf_1061316296_noimgclip_dft_averemove_swbh_dencorr_2dkpower_orig

And after the bug fix:
fhd_nb_model_gaussdecomp_update_nofreqdep_corrmin_double_uvf_1061316296_noimgclip_dft_averemove_swbh_dencorr_2dkpower

bhazelton:
That oscillatory pattern in the pre-fix spectra is fascinating. I'm a little disturbed by the diagonal line in the window in the post-fix spectra. Is that a known artifact? It looks like something Adam found years ago, but I thought we'd killed that one.

nicholebarry:
You are 100% correct! It's the "ghost line," come back to haunt us again. 

Over the years we've backed off the extreme overresolution of the beam because of the smart beam interpolation code that Ian wrote. But it looks like I've found the floor again. Increasing the resolution of the beam moves the line about, as expected.

bhazelton:
Fascinating! I guess I should be glad, it means the floor is dropping :-P.

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -187,7 +190,7 @@ struct={code_version:String(code_version),instrument:String(instrument),obsname:
     phasera:meta.phasera,phasedec:meta.phasedec,orig_phasera:meta.orig_phasera,orig_phasedec:meta.orig_phasedec,$
     n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:n_time_cut,$
     n_vis:Long(n_vis),n_vis_in:Long(n_vis_in),n_vis_raw:Long(n_vis_raw),nf_vis:Long(n_vis_arr),primary_beam_area:Ptrarr(4),primary_beam_sq_area:Ptrarr(4),pol_names:pol_names,$
-    jd0:meta.jd0,max_baseline:Float(max_baseline),min_baseline:Float(min_baseline),delays:meta.delays,lon:meta.lon,lat:meta.lat,alt:meta.alt,$
+    jd0:meta.jd0,max_baseline:max_baseline,min_baseline:min_baseline,delays:meta.delays,lon:meta.lon,lat:meta.lat,alt:meta.alt,$
```
isullivan, line 193: If min and max baseline are sometimes going to be doubles, then we need to force them to always be doubles or else the obs structure will sometimes be unreadable.

fhd_core/setup_metadata/fhd_struct_update_obs.pro
```diff
@@ -50,8 +50,8 @@ struct.dimension=Float(dimension)
 struct.elements=Float(elements)
 struct.kpix=Float(kbinsize)
 struct.degpix=Float(degpix)
-struct.max_baseline=Float(max_baseline)
-struct.min_baseline=Float(min_baseline)
+struct.max_baseline=max_baseline
+struct.min_baseline=min_baseline
```
isullivan, line 54: As above, if these are allowed to sometimes be doubles, we have to force them to always be doubles.

isullivan (approved):
Please note that obs structure parameters must be defined consistently, so if a parameter sometimes needs to be a `Double` it must always be a `Double`. Also, I see you are changing `visibility_degrid.pro` but not `visibility_grid.pro`. Could you please check whether the latter also needs modification?

isullivan (approved):

PR #249, opened 2021-01-07
Check if metadata flags all tiles

nicholebarry:
Instead of throwing an error after calibration/modelling, throw an error as soon as it's apparent that the metadata has flagged all of the tiles. 

(In the MWA, metadata flags all of the tiles if it did not get reporting stats from the receiver, thus being unable to determine the quality of the array during the obs).


rlbyrne (approved):
This looks good! Very useful change, thank you.

Issue #176, opened 2019-08-14
vis averaging bug


nicholebarry:
In vis_average.pro, n_freq is undefined if just vis_time_average is used.

nicholebarry:
Also, the params structure is missing antenna1 and antenna2

PR #253, opened 2021-02-18
Vis avg update

nicholebarry:
Simple update of vis averaging code, defining n_freq and the antenna structures. Solves #176


rlbyrne (approved):
Looks good!

PR #256, opened 2021-04-07
Further memory reduction

nicholebarry:
Implementing strategies used in the past to the general antenna response.

Instead of calculating the general antenna response at every location in the full zero-padded beam image, only calculate it at pixel locations within the horizon (i.e. where it is non-zero). This is a drastic improvement in memory (reducing from 70Gb required to 20Gb on my current test, reduction will be dependent on beam resolution and zero-padding) and a good improvement in time (no reforming required, and no calculating 0 many many times, reduction by a few minutes).

### DIFF
--- a/fhd_core/beam_modeling/general_antenna_response.pro
+++ b/fhd_core/beam_modeling/general_antenna_response.pro
@@ -11,14 +11,10 @@ response=Ptrarr(n_ant_pol,n_ant)
 nfreq_bin=antenna[0].nfreq_bin
 freq_center=antenna[0].freq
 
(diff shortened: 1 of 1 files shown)


nicholebarry:
No changes in PS. I did note, however, that the mem reduction is heavily based on what type of core it runs on. I think it has to do with how fast memory can be released.

mwilensky768 (approved):
It seems like this just threads `pix_use` into the calculation a little earlier so that the ensuing matrix/array multiplies after the projection calculation are smaller. This has the added bonus of flattening some of the arrays implicitly, reducing the number of explicit calls to `reform`. 

I'm pretty sure I follow the flattening math, and it makes me extra confident that the PS comes out the same, so I'll merge.

mwilensky768:
And I suppose it's only fitting that the 2^8 PR is a memory reduction.

PR #257, opened 2021-04-07
Fix box matrix bug and add gridding comments

nicholebarry:
In a rarely used option, there was a bug in where the box matrix was being initialized. The value of vis_n changes in this rarely used option, so the box matrix should be initialized right before it is used instead of outside the loop.

I then got carried away and started putting in comments in the gridding function. Help would most definitely be appreciated!


isullivan (dismissed):
Thank you for adding and improving the code comments.

The code change for the `box_matrix` definition looks fine to me.

isullivan (approved):
Looks good, please squash the commits and merge.

Issue #258, opened 2021-04-12
Minimum number of baselines and the Kalman filter


nicholebarry:
If the min_cal_solutions is not met for a tile per a freq, then a gain is not solved for that tile for that freq. This breaks the Kalman filter, because it tries to fit the 0 and the convergence tests are not set up for the special case of 0. 

Do we then flag that tile for that frequency if it doesn't satisfy the minimum number of baselines needed for a solution? If so, we should exclude that point from the Kalman filter.

nicholebarry:
Solved with pull request #259

PR #259, opened 2021-04-14
Remove gains which don't satisfy conditions from strict convergence

nicholebarry:
Preemptively set gains which do not satisfy the minimum number of equations set forth in the linear least squares solver to 0. This removes them from the strict convergence threshold, solving issue #258. Otherwise, that frequency generally fails the strict convergence rather quickly. 

It should be noted that this biases the loose convergence threshold (a mean of the tiles for the frequency). However, this was present before as well for other zeroed cases.


mwilensky768 (approved):
@nicholebarry and I had a conversation about this offline. Just giving another summary of the problem/solution, and my approval.

Some tiles do not satisfy the minimum number of cal solutions required to solve for a gain in a well-defined way at some frequency/pol. If these are not pre-emptively set to 0, then the calibration routine tries to fit that zero, and results in a small number. 

This feeds into the strict calibration threshold evaluation since this asks about a fractional difference between the old and new gain. When the old gain is small, there is a small number in the denominator, causing a huge number far above the convergence threshold. If this is set to 0, as the changes do, the logic of weight_invert causes the strict convergence criterion to be exactly 0, so that it does not fail the strict convergence test.

There are a couple comments in the code referring to this along with the code changes.

PR #261, opened 2021-05-20
Correct crosspol phase sign error

### LINKED ISSUE #260: Possible bug in crosspol phase calculation
I believe the calculation in `vis_calibration_crosspol_phase` is incorrect. Here's a writeup of what I think the corrected math would be:

[Potential_Crosspol_Phase_Calculation_Error.pdf](https://github.com/EoRImaging/FHD/files/6475202/Potential_Crosspol_Phase_Calculation_Error.pdf)

Testing is needed to see if this is a real bug and to determine how large of an effect it is.

rlbyrne:
Closes #260


rlbyrne:
I tested this change on 5 observations near EoR-0. The new calculation appreciably changes the values of the fit crosspol phase. Here are the before and after phases for each of the 5 observations, in radians:

Screen Shot 2021-05-20 at 1 29 05 PM

The new calculation reduces structure in Stokes V, presumably by reducing Stokes U to V leakage. Here is an example of one observation with the old calculation:
1131454296_Residual_V_old
and with the new calculation:
1131454296_Residual_V_new
Here is the differenced image, old-new:
1131454296_Residual_V_diff

PR #263, opened 2021-06-01
Nan flux hotfix

nicholebarry:
If nan fluxes are present, then warn the user and remove them from source modelling.


rlbyrne (approved):
This change is tested and works

PR #270, opened 2021-08-27
Model transfers

nicholebarry:
Extra instances to transfer 1) model visibilities and 2) a simulated model uv-plane without the instrument. This allows for the code to input model visibilities in a more efficient manner (i.e. in a separate transfer module), or to input a theoretical model uv-plane with only source contributions. The latter is helpful for simulations/parameter changes, and the former is helpful for double data runs (separate kernel for cal and gridding).


dstorer (approved):
Looks good to me - this will be super useful, so thanks for doing it!

PR #271, opened 2021-09-30
Adding robust filter

nicholebarry:
This is an implementation of robust/Briggs weighting for images, where `robust_param` is used to encode the robust scaling. It ranges from -2 to 2, where -2 is close to uniform weighting and 2 is close to natural weighting. Below are simulations of point sources in dirty Stokes I for robust weightings of -2, -1, 0, 1, and 2 (no noise).

Here is a locked DS9 tiled image:
Screen Shot 2021-09-30 at 2 14 04 pm
And again, oversaturated
Screen Shot 2021-09-30 at 2 13 33 pm
And a gif of the automatically made FHD images
image

The extreme ends look correct (-2 ~ uniform, 2 ~ natural), but I have no frame of reference for in-between values. It would be great if someone experienced in what robust weighting looks like could advise if this is the normal behaviour.


bhazelton (approved):
This looks reasonable to me. I'm really not an expert on this, but it looks like it won't mess anything else up and is just for the output image weighting, so unlikely to have other consequences.

PR #272, opened 2021-10-07
Passing variables into filter functions

nicholebarry:
If `obs`,`psf`, and `params` are passed into `filter_uv_uniform`, then the visibility count needed for weighting is calculated on the fly. If not, then the `vis_count` file is restored. However, the above structs were not passed into the filter function in `dirty_image_generate`. 

I may see why...the filter function call in `dirty_image_generate` is general, i.e. it can call any filter function depending on input. Some of the simpler filter functions don't need `obs`,`psf`, and `params` to calculate their filter. 

I think that it's better to not rely on I/O, so I've added the structs as passable variables in each filter function, even if they aren't explicitly needed, just so the general filter function call won't fail in `dirty_image_generate`. This is a matter of opinion, so please decline if you don't think it's worth it.

### DIFF
--- a/fhd_output/fft_filters/filter_uv_hanning.pro
+++ b/fhd_output/fft_filters/filter_uv_hanning.pro
@@ -1,4 +1,5 @@
-FUNCTION filter_uv_hanning,image_uv,name=name,weights=weights,filter=filter,return_name_only=return_name_only,_Extra=extra
+FUNCTION filter_uv_hanning,image_uv,obs=obs,psf=psf,params=params,name=name,$
+    weights=weights,filter=filter,return_name_only=return_name_only,_Extra=extra
 name='hanning'
 IF Keyword_Set(return_name_only) THEN RETURN,image_uv
 IF N_Elements(weights) NE N_Elements(image_uv) THEN RETURN,image_uv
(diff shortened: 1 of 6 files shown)


fhd_output/fft_filters/filter_uv_hanning.pro
```diff
@@ -1,4 +1,5 @@
-FUNCTION filter_uv_hanning,image_uv,name=name,weights=weights,filter=filter,return_name_only=return_name_only,_Extra=extra
+FUNCTION filter_uv_hanning,image_uv,obs=obs,psf=psf,params=params,name=name,$
+    weights=weights,filter=filter,return_name_only=return_name_only,_Extra=extra
```
isullivan, line 2: I think you don't have to change the call signature, since if `obs`, `psf`, and `params` are passed in they will get bundled in `extra`.

fhd_output/fft_filters/filter_uv_natural.pro
```diff
@@ -1,5 +1,5 @@
-FUNCTION filter_uv_natural,image_uv,name=name,weights=weights,filter=filter,$
-    return_name_only=return_name_only,_Extra=extra
+FUNCTION filter_uv_natural,image_uv,obs=obs,psf=psf,params=params,name=name,weights=weights,$
+    filter=filter,return_name_only=return_name_only,_Extra=extra
```
isullivan, line 2: As above, I think `extra` will catch any extra parameters that are supplied.

fhd_output/fft_filters/filter_uv_radial.pro
```diff
@@ -1,5 +1,5 @@
-FUNCTION filter_uv_radial,image_uv,name=name,weights=weights,filter=filter,$
-    radial_power=radial_power,return_name_only=return_name_only,_Extra=extra
+FUNCTION filter_uv_radial,image_uv,obs=obs,psf=psf,params=params,name=name,weights=weights,$
+    filter=filter,radial_power=radial_power,return_name_only=return_name_only,_Extra=extra
```
isullivan, line 2: As above, `extra` should swallow the extra parameters.

isullivan (approved):
This looks fine to me now, but I recall I avoided this approach on purpose back when it was written. I think the problem is that there are instances when we call `dirty_image_generate` when the required structures (`params` especially) are not available, so instead in the few times that those are needed we fetch them from disk. So, my inclination would be to leave the code as-is since I'm not aware of it currently causing problems, but I have no major objection.

nicholebarry:
I agree... let's not merge this.

PR #274, opened 2021-10-13
Redundant code to a single file

nicholebarry:
There was a lot of redundant code between `visibility_grid`,`visibility_degrid`, and `vis_count`. I've condensed the majority of the redundant code to a new file, `baseline_grid_locations`. This file returns information regarding the locations of the unflagged visibilities in units of pixels in a variety of different forms.

Each of the three original files accounted for flagging in a slightly different way, and I've recreated each of those. The only exception being that unsetting `fill_model_visibilities` during the model making process will emulate the flagging done for gridding with this new file -- arguably what we want to do anyways. In any case, `fill_model_visibilities` is currently hard set on anyways. 

The power spectra for a typical run are the same between the old branch and the new branch. The images, however, are different on extremely low levels (i.e. 10^-20), which makes me think I haven't fully replicated some float/double behaviour during image making. I will continue to investigate this, but I'd like to get this branch reviewed as fast as possible so that it can be ported to python via Joel in the next 2 weeks.

### DIFF
--- a/fhd_core/gridding/baseline_grid_locations.pro
+++ b/fhd_core/gridding/baseline_grid_locations.pro
@@ -7,7 +7,8 @@
 pro baseline_grid_locations,obs,psf,params,xmin=xmin,ymin=ymin,vis_weight_ptr=vis_weight_ptr,$
   fill_model_visibilities=fill_model_visibilities,bi_use=bi_use,fi_use=fi_use,$
   vis_inds_use=vis_inds_use,interp_flag=interp_flag,dx0dy0_arr=dx0dy0_arr,dx0dy1_arr=dx0dy1_arr,$
-  dx1dy0_arr=dx1dy0_arr,dx1dy1_arr=dx1dy1_arr,mask_mirror_indices=mask_mirror_indices
+  dx1dy0_arr=dx1dy0_arr,dx1dy1_arr=dx1dy1_arr,x_offset=x_offset,y_offset=y_offset,$
+  mask_mirror_indices=mask_mirror_indices
 
   ; Extract information from the structures
   n_tile=obs.n_tile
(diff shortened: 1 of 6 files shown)


fhd_core/gridding/visibility_count.pro
```diff
@@ -1,106 +1,47 @@
...
-;match all visibilities that map from and to exactly the same pixels
-bin_n=histogram(xmin+ymin*dimension,binsize=1,reverse_indices=ri,min=0) ;should miss any (xmin,ymin)=(-1,-1) from weights
+; Match all visibilities that map from and to exactly the same pixels and store them as a histogram in bin_n
+; with their respective index ri. Setting min equal to 0 excludes flagged (i.e. (xmin,ymin)=(-1,-1)) data
+bin_n=histogram(xmin+ymin*dimension,binsize=1,reverse_indices=ri,min=0)
 bin_i=where(bin_n,n_bin_use)
 
-weights=fltarr(dimension,elements)
+double_precision=0
+IF Tag_Exist(obs, 'double_precision') THEN double_precision=obs.double_precision
```
isullivan, line 31: This can just be `double_precision=obs.double_precision`

fhd_core/gridding/baseline_grid_locations.pro
```diff
@@ -0,0 +1,158 @@
...
+
+  IF Keyword_Set(mask_mirror_indices) THEN BEGIN
+    ; Option to exlude v-axis mirrored baselines
+    IF n_conj GT 0 THEN BEGIN
+      xmin[*,conj_i]=-1
+      ymin[*,conj_i]=-1
+    ENDIF
+  ENDIF
+
+END
```
isullivan, line 158: You can also include the following code here:
`; Match all visibilities that map from and to exactly the same pixels and store them as a histogram in bin_n
; with their respective index ri. Setting min equal to 0 excludes flagged (i.e. (xmin,ymin)=(-1,-1)) data
bin_n=histogram(xmin+ymin*dimension,binsize=1,reverse_indices=ri,min=0)`

I think it might make sense to turn `baseline_grid_locations` into a function and return `bin_n`, but I don't feel strongly about that.

nicholebarry:
I've added your suggestions, and the power spectrum differences are still zero. Good to go!

PR #277, opened 2021-11-25
Beam cleanup

nicholebarry:
There were a few bugs with the gaussian decomp branch, particularly on the memory-restricted loop and the beam image generation. 

The psf transfer was also very opaque. This version is easier to read and has better error messaging.


rlbyrne (approved):
I've tested this branch with a basic FHD run (without using the Gaussian beam decomposition) and it works.

PR #281, opened 2022-01-17
Mem splitting bug fix

nicholebarry:
I thought I had fixed this, but apparently I hadn't. 

It's a splitting of the array into chunks if a lower memory footprint is required. Tested and works now. Only affects the Gaussian decomposition.


dstorer (approved):
Looks good to me!

PR #282, opened 2022-01-20
group_arr psf.id bug

nicholebarry:
As mentioned in Slack:

> I think Joel may have found a bug, and I wanted to get your opinion.
> Line 64 of visibility_grid.pro finds the group identifier if there are multiple beams being used:
> group_arr=reform(psf.id[polarization,freq_bin_i[fi_use],bi_use])
> However, bi_use has the same dimensionality as the num of visiblities (4 million something), and psf.id has the dimensionality of x, x, 8128 . It doesn’t error, but I think the group_arr is set to the last value in the array if it exceeds 8128.
> I think maybe bi_use should be bi_use_reduced. I wanted to check with you though, because it seemed like this could have potential for IDL black magic.
> Maybe this hasn’t been an issue thus far because we generally use one beam. And maybe this explains why our dead dipole tests never really went anywhere.

I played around with a couple of iterations, and this was the fastest/most compact. Still looks a little clunky, so I'm open to suggestions.

### DIFF
--- a/fhd_core/gridding/visibility_grid.pro
+++ b/fhd_core/gridding/visibility_grid.pro
@@ -61,7 +61,6 @@ frequency_array=frequency_array[fi_use]
 complex_flag=psf.complex_flag
 psf_dim=psf.dim
 psf_resolution=psf.resolution
-group_arr=reform(psf.id[polarization,freq_bin_i[fi_use],bi_use])
 beam_arr=*psf.beam_ptr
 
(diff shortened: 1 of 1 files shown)


isullivan (approved):
Assuming this gives the same results for our standard runs with homogenous arrays, this looks like a decent fix to me.

isullivan (approved):
This should be fine for our standard case when the antennas are homogenous, and it ought to fix this bug.

PR #283, opened 2022-03-18
Multi-beam fixes

nicholebarry:
There were a few lingering bugs regarding the ID group array, which determines which baselines have the same beams. This is a rarely used option (at this stage).

This has been tested to give the proper IDs on two separate tests. There may remain some edge cases that I wasn't able to think of, but this should cover most cases with this option.

A warning was also added from the suggestion of Joel.


isullivan (approved):
Seems OK

PR #285, opened 2022-04-07
antenna indices rewrite

### LINKED ISSUE #292: calculation of n_tile and antenna number handling
The number of tiles `n_tile` is calculated as the largest antenna number present. When uvfits files are submitted with actual antenna numbers rather than indices, the number of tiles FHD assumes can be arbitrarily large. This becomes problematic for MWA data in particular, as the number of antennas in the metafits file is 128 and will not match `n_tile`.

nicholebarry:
Quick little fix for the pyuvdata antenna naming convention. I admit that it may not be the best fix....

For starters, there is an assumption that the number of tiles would be less than antenna names, and that the antenna names are numbers. These are both assumptions that could break. 

fixes #292

### DIFF
--- a/fhd_core/setup_metadata/fhd_struct_init_obs.pro
+++ b/fhd_core/setup_metadata/fhd_struct_init_obs.pro
@@ -134,9 +134,9 @@ if max(params.antenna1) GT n_tile then begin
     tile_B = params.antenna2
     for tile_i=0, n_tile-1 do begin
         inds = where(layout.antenna_numbers[tile_i] EQ (params.antenna1),n_count)
-        if n_count GT 0 then tile_A[inds] = tile_i+1
+        if n_count GT 0 then tile_A[inds] = tile_i+1 else message, "antenna numbers in uvfits vis and ant hdus do not match"
         inds = where(layout.antenna_numbers[tile_i] EQ (params.antenna2),n_count)
-        if n_count GT 0 then tile_B[inds] = tile_i+1
+        if n_count GT 0 then tile_B[inds] = tile_i+1 else message, "antenna numbers in uvfits vis and ant hdus do not match"
     endfor
     params.antenna1 = tile_A
     params.antenna2 = tile_B
--- a/fhd_core/setup_metadata/fhd_struct_init_obs.pro
+++ b/fhd_core/setup_metadata/fhd_struct_init_obs.pro
@@ -59,17 +59,13 @@ ENDIF ELSE BEGIN
     grid_info=Ptr_new()
 ENDELSE
 
-antenna_flag = 1
 IF Tag_exist(params, "antenna1") AND Tag_exist(params, "antenna2") THEN BEGIN
     ant1_arr = params.antenna1
     ant2_arr = params.antenna2
-    IF Max(ant1_arr) GT 0 AND Max(ant2_arr) GT 0 THEN BEGIN
-        tile_A = ant1_arr
-        tile_B = ant2_arr
-        n_tile=max(tile_A)>max(tile_B)
-        antenna_flag = 0
-    ENDIF
-ENDIF
+    tile_A = ant1_arr
+    tile_B = ant2_arr
+    antenna_flag=0 
+ENDIF ELSE antenna_flag = 1
 IF antenna_flag THEN BEGIN
     ;256 tile upper limit is hard-coded in CASA format
     ;these tile numbers have been verified to be correct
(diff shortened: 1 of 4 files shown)


PyxieLouStar:
fixes #292

PyxieLouStar:
@dstorer will test this branch on a HERA file.

bhazelton (commented):
I can't comment on the FHD structure side, but this looks correct to me in terms of the way it's using the information in the uvfits file.

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -132,7 +128,21 @@ IF N_Elements(min_baseline) EQ 0 THEN min_baseline=Min(kr_arr[where(kr_arr)]) EL
...
-meta=fhd_struct_init_meta(file_path_vis,hdr,params,degpix=degpix,dimension=dimension,elements=elements,$
+; If it appears that the params struct uses tile names instead of indices, then rewrite with indices
+if max(params.antenna1) GT n_tile then begin
+    tile_A = params.antenna1
+    tile_B = params.antenna2
+    for tile_i=0, n_tile-1 do begin
+        inds = where(layout.antenna_numbers[tile_i] EQ (params.antenna1),n_count)
+        if n_count GT 0 then tile_A[inds] = tile_i+1
+        inds = where(layout.antenna_numbers[tile_i] EQ (params.antenna2),n_count)
+        if n_count GT 0 then tile_B[inds] = tile_i+1
```
isullivan, line 139: Should you set `tile_A` and `tile_B` differently if `n_count` is ever 0? I could imagine a situation where a tile was flagged and possibly not in `layout.antenna_numbers`, and it could either have the same value of `params.antenna1` (or `2`) as an existing tile index, or it could be greater than `n_tile`. Either way could lead to errors. One option could also be to initialize `tile_A` and `tile_B` to something like -1, and print a warning if there are any -1's remaining after this tile number reassignment.

fhd_core/visibility_manipulation/uvfits_read.pro
```diff
@@ -84,7 +84,21 @@ ENDIF ELSE BEGIN
     ;free memory
     data_array=0 
     vis_weights0=0
-    
+
+    ;; now read extensions one by one until we find the antenna table
+    t_read_ant=Systime(1)
+    ext_name = ''
+    while ext_name ne 'AIPS AN' do begin
+        ext_data = mrdfits(lun, 0, ext_header)
```
isullivan, line 92: I'm a little rusty on IDL's `mrdfits()` function, but won't this always read the first extension?

isullivan (approved):
Looks OK to me, and glad to hear it fixes the problems with antenna numbers. Please see my (minor) comments.

fhd_core/visibility_manipulation/uvfits_read.pro
PyxieLouStar: This code wasn't changed in this branch, so I didn't test it. But from reading the `mrdfits()` documentation, I think that when the function is given an open file, it will step through the extensions one-by-one. If a nonzero number is given as the second argument, it will skip that number of extensions and read in the next one.

fhd_core/setup_metadata/fhd_struct_init_obs.pro
PyxieLouStar: If `n_count` is zero, then the submitted uvfits file is broken. `params.antenna1/2` are set from the uvfits visibility extension baseline or antenna arrays. These arrays can only have entries which are also in the antenna extension antenna array, which is what `layout.antenna_numbers` comes from. So maybe we should throw an error if `n_count` is zero?

isullivan: Yes, good idea to throw an error if that means the file is broken.

PyxieLouStar: actually I have that wrong. This is iterating over `layout.antenna_numbers`; it is possible for an antenna to be present in `layout.antenna_numbers` and not be present in `params.antenna1/2`. So if `n_count` is zero, that antenna is not present in `params.antenna1/2` and thus in not present in `tile_A/B`. No extra handling is necessary if `n_count` is zero.

isullivan (approved):
It looks like the last two commits add and then remove an error message. Since these cancel each other, could you remove these commits? It's probably not a bad idea to have an error message in this case, though, so I would suggest adding a new error message.

PR #288, opened 2022-04-22
Cleanup of in situ inputs and spacing

nicholebarry:
Cleanup of input keywords and spacing of the in situ input. Did a quick run through with a small sim to make sure there was no drastic failures.


mwilensky768 (approved):
My understanding of the changes are as follows:

(1) Some extraneous keywords were being passed in fhd_main to in_situ_sim_setup related to calibration and a couple other things not actually used by the sim wrapper (which I also checked), so these were just removed from the call to the sim wrapper.

(2) The keywords removed from the call in (1) were also removed on this end. Most of the rest of the changes solely enhance readability. One final change is that a couple temporary variables are renamed to reflect as much.

In summary, the changes are transparent, and highly unlikely to change the functionality of the code by my account. Additionally, since @nicholebarry tested the wrapper, I'm very confident these changes can be merged.

PR #289, opened 2022-04-22
Remove nans for cal plotting residuals

nicholebarry:
Untested. This should allow for the cal residual plots to be made even if there are nans in the gain solutions (i.e. flagged channels).


PyxieLouStar:
I tested using obs 1321445344 which had not been producing residual plots due to NANs in the data. Looks good!

PyxieLouStar (approved):

PR #290, opened 2022-04-26
Diffuse options

nicholebarry:
In order to reduce aliasing when FFTing the diffuse map, I've added options for anti-aliasing windows before and after the FFT. I also found a small error with reading in Kelvin maps, which I think made it really difficult to compare Jy/str maps with Kelvin maps.

----Bug fix

When calculating the pixel area, the nside of the Healpix map was used instead of the pixel size of the flat, output map. Since the calculation is used to generate an output Jy/pixel map, the pixel area needs to be the output size. This doesn't affect Jy/str maps. This was in `healpix_interpolate.pro` which doesn't seem to be used anywhere else, so I think it's safe to change it there.

----Diffuse options

This gives the user the option to play with a variety of different spectral windows. However, I've found the following to be the best balance between using the majority of the diffuse map whilst reducing systematics. If more of the diffuse map is removed, then the calibration starts to deteriorate. 

Pre-FFT: Tukey window, where the alpha is calculated to align with the primary lobe. This allows the most instrumentally sensitive area of the diffuse map to remain to maximise calibration ability while still having a taper. Specifically, the taper begins at 1% beam level.
Post-FFT: A Blackman-Harris window. This keeps most everything under 50 wavelengths. I've also put in another Tukey option that tapers after 50 wavelengths (for the more conservative types).

### DIFF
--- a/fhd_core/HEALPix/healpix_interpolate.pro
+++ b/fhd_core/HEALPix/healpix_interpolate.pro
@@ -48,7 +48,7 @@ image_mask[Min(Floor(xv_hpx)):Max(Ceil(xv_hpx)),Min(Floor(yv_hpx)):Max(Ceil(yv_h
 
 x_frac=1.-(xv_hpx-Floor(xv_hpx))
 y_frac=1.-(yv_hpx-Floor(yv_hpx))
-IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,nside=nside,freq=obs.freq_center) $
+IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,degpix=obs.degpix,freq=obs.freq_center) $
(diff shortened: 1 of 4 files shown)


fhd_core/HEALPix/healpix_interpolate.pro
```diff
@@ -48,7 +48,7 @@ image_mask[Min(Floor(xv_hpx)):Max(Ceil(xv_hpx)),Min(Floor(yv_hpx)):Max(Ceil(yv_h
 
 x_frac=1.-(xv_hpx-Floor(xv_hpx))
 y_frac=1.-(yv_hpx-Floor(yv_hpx))
-IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,nside=nside,freq=obs.freq_center) $
+IF Keyword_Set(from_kelvin) THEN pixel_area_cnv=convert_kelvin_jansky(1.,degpix=obs.degpix,freq=obs.freq_center) $
```
rlbyrne, line 51: I just tested this change with LWA data and the LWA m-mode map and confirmed this bug fix. This is confusing syntax for this function, though. We may want to change.

rlbyrne (approved):

PR #291, opened 2022-05-17
Gaussian model update

nicholebarry:
I updated the gaussian source model generation. There are a couple of changes:

1) better gaussian parameter passing. I created a new structure so things are now much more compact and easier to manage.

2) including image plane -> uvw compression math from Jaiden's upcoming paper. Instead of calculating the mean compression from output from `apply_astrometry`, I'm calculating the expected compression using geometry. I've attached a screenshot of the relevant math. 

Screen Shot 2022-05-16 at 10 54 03 am

3) switching the `source_dft` matrix ordering so that it's column major to be faster in IDL. This was a simple fix that significantly decreased walltime and mem requirements.

4) small bugs and general spacing. 

The main branch (with Christene's LoBES catalog) took 381 min at 94 Gb for 69 118 sources. The new branch took 326 min at 65 Gb for the same sources. Running the main branch and the new branch with just point sources yielded the exact same power spectra. 

Here is the difference between the old branch and the new branch (blue means less power in the new branch).
fhd_nb_data_pointing_beam_lobes_update_int_fullimg_dft_averemove_swbh_dencorr__oldbranch_1061316296_minus_newbranch_FoV_1061316296_2dkpower

And here is the ratio difference between the old branch and the new branch (blue means more power was removed in the new branch.
fhd_nb_data_pointing_beam_lobes_update_int_fullimg_dft_averemove_swbh_dencorr__oldbranch_1061316296_diffratio_newbranch_FoV_1061316296_2dkpower

The difference suggests that the geometrical argument behaves better than the mean calculation as projection effects increase towards the horizon.

### DIFF
--- a/fhd_core/source_modeling/source_dft_multi.pro
+++ b/fhd_core/source_modeling/source_dft_multi.pro
@@ -52,42 +52,41 @@ FOR a_i=0L,n_alpha-1 DO BEGIN
 ENDFOR
 
 IF keyword_set(gaussian_source_models) then begin
-  if tag_exist(source_array, 'shape') then begin
-    gauss_test = ceil(0>(source_array.shape.x + source_array.shape.y)<1)
-    gauss_inds = where(gauss_test GT 0,n_gauss)
-  endif else n_gauss=0
+    if tag_exist(source_array, 'shape') then begin
+        gauss_test = ceil(0>(source_array.shape.x + source_array.shape.y)<1)
+        gauss_inds = where(gauss_test GT 0,n_gauss)
+    endif else n_gauss=0
 
-  if n_gauss GT 0 then begin
-      ;Convert from FWHM in arcsec to stddev in deg
-      gaussian_ra=double(source_array[gauss_inds].shape.x)/(7200*sqrt(2*alog(2)))
-      gaussian_dec=double(source_array[gauss_inds].shape.y)/(7200*sqrt(2*alog(2)))
-      ;Convert from deg to rad
-      gaussian_rot=double(source_array[gauss_inds].shape.angle)*!DPi/180.
+    if n_gauss GT 0 then begin
+        ;Convert from FWHM in arcsec to stddev in deg
+        gaussian_ra=double(source_array[gauss_inds].shape.x)/(7200*sqrt(2*alog(2)))
+        gaussian_dec=double(source_array[gauss_inds].shape.y)/(7200*sqrt(2*alog(2)))
+        ;Convert from deg to rad
+        gaussian_rot=double(source_array[gauss_inds].shape.angle)*!DPi/180.
 
-      ;Compression due to flat sky approximation of curved sky, see J Cook et al. 2022
-      Eq2Hor,source_array[gauss_inds].ra,source_array[gauss_inds].dec, obs.JD0, gauss_alt, gauss_az, nutate=1,precess=1,$
-        aberration=0, refract=0, lon=obs.lon, alt=obs.alt, lat=obs.lat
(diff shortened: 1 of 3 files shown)


bhazelton:
Those power spectra look really promising!

fhd_core/source_modeling/source_dft_multi.pro
```diff
@@ -50,78 +52,43 @@ FOR a_i=0L,n_alpha-1 DO BEGIN
...
+        gaussian_y = (gaussian_y_vals[*,1]-gaussian_y_vals[*,0]) / gaussian_dec_corr
 
-      ;Gaussian source model major and minor axes RA DEC
-      ;           _ a_2,d_2
-      ;          / \
-      ; a_3,d_3 |   | a_1,d_1
-      ;          \_/
-      ;   a_4,d_4
-      ;
-      ;Convert from deg to pixels
```
isullivan, line 83: Is this diagram no longer relevant? It would be helpful if you could update it rather than delete it if that is possible.

fhd_core/source_modeling/source_dft_multi.pro
```diff
@@ -50,78 +52,43 @@ FOR a_i=0L,n_alpha-1 DO BEGIN
...
-        gaussian_dec[source_ind]=0
-        gaussian_rot[source_ind]=0
-      endelse
-    endfor
-    inds_gauss=where(gaussian_ra,n_gauss_params,complement=inds_point)
-    if n_gauss_params eq 0 then begin
-      print, 'Catalog does not contain Gaussian shape parameters. Unsetting keyword gaussian_source_models.'
-      undefine, gaussian_source_models
-    endif else begin
+        gaussian_rot=double(source_array[gauss_inds].shape.angle)*!DPi/180.
```
isullivan, line 65: Here and throughout this file, please use `!Radeg` instead of `!DPi/180`

fhd_utils/FFT/source_dft.pro
```diff
@@ -47,149 +46,164 @@ x_use=x_loc_use-dimension/2.
...
-            ENDIF
+    ;Special case of memory bins not being able to hold needed binsize for all DFTs
+    if memory_bins*sources_per_bin LT n0 then begin
+        unbinned_sources=n0-Total(binsize) ;Number of leftover sources that can't held currently
+        num_more_bins=Ceil(unbinned_sources/sources_per_bin) ;Number of bins that need to be added to hold all sources at required binsize
+        memory_bins+=num_more_bins ;Recalculate number of bins necessary
+        binsize=Lonarr(memory_bins)+sources_per_bin ;Recalculate binsize given new memory bins
+    endif
+
+    ;Make the last bin forces number of sources distributed throughout bins equals the total num of sources.
```
isullivan, line 93: Minor, but could you re-write this description?

isullivan (dismissed):
These changes look good, especially with the apparent improvement in the power spectrum. I had a couple very minor comments on the code changes.

fhd_core/source_modeling/source_dft_multi.pro
nicholebarry: The definition of the variables are not longer relevant because I removed the corresponding code. Bit unfortunate, really.

fhd_core/source_modeling/source_dft_multi.pro
nicholebarry: Unfortunately, `!Radeg=57.295776` is floating precision, so I would rather stay with `!Dpi/180=57.295779513082323`.

isullivan (approved):
Thank you for the clarifications on my questions before. Everything looks good to merge.

PR #297, opened 2022-11-29
Fix bugs causing errors while gridding/degridding

PyxieLouStar:
Some uvfits files were resulting in the following error during gridding/degridding:
`% Unable to dereference NULL pointer: <POINTER  (<NullPointer>)>.`

This was tracked down to two different bugs. In one case, the code which maps antenna numbers to indices was not being triggered. In another case, only using one of the two uvfits date fields resulted in an incorrect calculation of the number of baselines.

Both of these bugs are fixed here. Additionally, I have added a check which makes sure that all the elements from the uvfits header antenna arrays are present in the antenna numbers array from the antenna table.

### DIFF
--- a/fhd_core/visibility_manipulation/vis_param_extract.pro
+++ b/fhd_core/visibility_manipulation/vis_param_extract.pro
@@ -4,7 +4,12 @@ FUNCTION vis_param_extract,params,header, antenna_mod_index=antenna_mod_index
 uu_arr=Double(reform(params[header.uu_i,*]))
 vv_arr=Double(reform(params[header.vv_i,*]))
 ww_arr=Double(reform(params[header.ww_i,*]))
-time=reform(params[header.date_i,*])
+
+; Use both date fields if available in the uvfits file
+; The center value is not added automatically, so add jd0
+IF n_elements(header.date_i) GT 1 THEN BEGIN
+    time=Double(reform(params[header.date_i[0],*])) + Double(reform(params[header.date_i[1],*])) + Double(header.jd0)
+ENDIF ELSE time=Double(reform(params[header.date_i,*]))
 
 ; Sort out the baseline numbering and antenna numbers
 IF header.baseline_i GE 0 THEN $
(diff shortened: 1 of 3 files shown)


fhd_core/visibility_manipulation/vis_param_extract.pro
```diff
@@ -4,7 +4,12 @@ FUNCTION vis_param_extract,params,header, antenna_mod_index=antenna_mod_index
 uu_arr=Double(reform(params[header.uu_i,*]))
 vv_arr=Double(reform(params[header.vv_i,*]))
 ww_arr=Double(reform(params[header.ww_i,*]))
-time=reform(params[header.date_i,*])
+
+; Use both date fields if available in the uvfits file
+; The center value is not added automatically, so add jd0
+IF n_elements(header.date_i) GT 1 THEN BEGIN
```
isullivan, line 10: Maybe use `EQ 2` instead of `GT 1`?

fhd_core/visibility_manipulation/vis_param_extract.pro
```diff
@@ -4,7 +4,12 @@ FUNCTION vis_param_extract,params,header, antenna_mod_index=antenna_mod_index
 uu_arr=Double(reform(params[header.uu_i,*]))
 vv_arr=Double(reform(params[header.vv_i,*]))
 ww_arr=Double(reform(params[header.ww_i,*]))
-time=reform(params[header.date_i,*])
+
+; Use both date fields if available in the uvfits file
+; The center value is not added automatically, so add jd0
```
isullivan, line 9: You added the comment about adding `jd0`, but then removed `jd0` in a later commit. It looks like the problem with the beam if `jd0` is included is that the double precision float overflows. That is probably why we avoided adding it in in the first place, so maybe all you need to do is update this comment.

isullivan (dismissed):
Looks fine, but please address my comment on the use of `jd0` in `vis_param_extract`.

fhd_core/visibility_manipulation/vis_param_extract.pro
PyxieLouStar: Looks like julian dates are actually calculated in `fhd_struct_init_meta`. I have updated the comment to reflect this.

isullivan (approved):

PR #302, opened 2023-03-29
Make transfer_psf work when beams and obs are in different directories

dstorer:
Minor update to allow `transfer_psf` keyword to work when the `beams` and `obs` structures are in different directories (as is true for the default output format). Previously, the if statement on line 24 would pass when `transfer_psf` was a directory, and then fail in `getvar_savefile` when trying to restore the directory - this change means that this if statement is skipped if the keyword is a directory rather than a file, and is instead handled in the following chunk of code.


PyxieLouStar (approved):
This looks good to me!

PR #310, opened 2023-04-25
Support data with more antennas than baselines

### LINKED ISSUE #309: beam_setup crashes when not all antenna pairs are represented in the data
I am processing a simulation in which a small subset of antenna pairs are actually represented as baseline measurements in the data set, so the number of antennas actually exceeds the number of baselines. However, the beam group syntax in `beam_setup` assumes that all possible antenna pairs are present in the data. This is the traceback of the error:
```
% Array used to subscript array contains out of range subscript: GROUP_ARR.
% Execution halted at: BEAM_SETUP        196
   /home/rbyrne/FHD/fhd_core/beam_modeling/beam_setup.pro
%                      FHD_MAIN           75
   /home/rbyrne/FHD/fhd_core/fhd_main.pro
%                      GENERAL_OBS       146
   /home/rbyrne/FHD/Observations/general_obs.pro
```

rlbyrne:
Closes #309


isullivan (approved):
It might be nice to ensure that the mod index is always a power of 2, but everything should still work like this.

PR #312, opened 2023-08-25
Fix indexing for `vis_baseline_hist`

SkyWa7ch3r:
This fixes an indexing issue found by Jack (@JLBLine) during the PyFHD testing, this should fix the results and save the values in each polarization rather than overwriting each polarization per loop inside the `vis_res_ratio_mean` and `vis_res_sigma` arrays. PyFHD and FHD get *almost* (precision error) the same results with this fix applied.

The previous `vis_baseline_hist`:

test_vis_baseline_hist_point_zenith_run1

The fixed `vis_baseline_hist`:

test_vis_baseline_hist_point_zenith_run1_after_fix


isullivan (approved):
Good catch, that was definitely a bug.

PR #314, opened 2023-10-26
quadratic interpolation of the beam in freq

nicholebarry:
Changing the default option on the interpolation on the beam in frequency from linear to quadratic to remove the coarse band lines in simulation.

Power_Spectrum_Plot


bhazelton (approved):
This is fantastic, thank you @nicholebarry !

PR #319, opened 2024-01-23
`apply_astrometry` refraction fix

SkyWa7ch3r:
I found a bug in `apply_astrometry` where refraction was applied by default when the intention was to ignore refraction by default. This means that `apply_astrometry` had different default behaviour to AstroPy which ignores refraction by default when doing any conversions unless both temperature and pressure are given. 

It's worth noting the refraction was being applied in world (RA/DEC) to pixel (x, y) and pixel (x, y) to world (RA/DEC) calculations where I doubt the atmosphere comes into play. So with that in mind the code I have done needs a review as these are not breaking changes, but may cause issues.

First thing I have done is changed the double negative of `/ignore_refraction` to a positive keyword `/refraction`. So in order to use refraction you MUST now explicitly use the `/refraction` keyword. This I believe makes the code easier to read and also makes your intentions clear to use `/refraction`.

With that change from a double negative keyword to a positive keyword, any calls to `apply_astrometry` had to change to keep FHD working as it did before. 

Anything that ***did not use*** the `/ignore_refraction` keyword now has the `/refraction` keyword ***applied***

i.e.
```
apply_astrometry, obs, ra_arr=ra, dec_arr=dec, x_arr=x, y_arr=y, /ad2xy
```

has been changed to

```
apply_astrometry, obs, ra_arr=ra, dec_arr=dec, x_arr=x, y_arr=y, /ad2xy, /refraction
```

Anything that ***did use*** the `/ignore_refraction` keyword has had the keyword removed and refraction is now ***ignored by default.***

i.e.

```
apply_astrometry, obs, ra_arr=ra, dec_arr=dec, x_arr=x, y_arr=y, /ad2xy, /ignore_refraction
```
has been changed to
```
apply_astrometry, obs, ra_arr=ra, dec_arr=dec, x_arr=x, y_arr=y, /ad2xy
```

The code needs to be checked as I believe there are likely instances where refraction was applied, but was likely not intended. If you made any of the code below, or extensively use the below functions, please check that the calls to `apply_astrometry` are doing what you intended for them to do.

### DIFF
--- a/fhd_utils/modified_astro/apply_astrometry.pro
+++ b/fhd_utils/modified_astro/apply_astrometry.pro
@@ -21,14 +21,14 @@
 ;    
 ;    ad2xy - set to convert from celestial to pixel coordinates
 ;    
-;    ignore_refraction - set to ignore refraction (for backwards compatibility and testing)
+;    refraction - set to take refraction into account (it's false by default for speed, backwards compatibility and testing)
 ;    
 ;    _Extra - passes any additional keywords (like temperature (Kelvin) or pressure (millibars) to co_refract
 ;
 ; :Author: Ian
 ;-
 PRO apply_astrometry,obs, x_arr=x_arr, y_arr=y_arr, ra_arr=ra_arr, dec_arr=dec_arr, xy2ad=xy2ad, ad2xy=ad2xy, $
-    astr=astr, JDate=JDate, ignore_refraction=ignore_refraction, _Extra=extra
+    astr=astr, JDate=JDate, refraction=refraction, _Extra=extra
 
 IF not (Keyword_Set(xy2ad) OR Keyword_Set(ad2xy)) THEN $
     message, "ERROR! At least one direction keyword (xy2ad or ad2xy) must be set."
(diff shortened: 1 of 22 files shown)


fhd_utils/modified_astro/apply_astrometry.pro
```diff
@@ -39,13 +39,16 @@ IF (Keyword_Set(xy2ad) AND Keyword_Set(ad2xy)) THEN $
 IF size(obs,/type) EQ 8 THEN BEGIN 
     astr=obs.astr
     JDate=obs.JD0
-ENDIF ELSE ignore_refraction=1
+ENDIF ELSE refraction=0
+
+if Keyword_Set(refraction) THEN refraction=1 ELSE refraction=0
```
isullivan, line 44: Minor point, but this is unnecessary since you only use `IF Keyword_Set(refraction)` below.

isullivan (approved):
Looks straightforward

nicholebarry:
Here is a PS test on one observation. Jackknife is a control run minus a run with the refraction changes. Changes are noise-like and small with no significant phase-space shape. WODEN models were transferred to both 
fhd_nb_noimgclip_dft_averemove_swbh7_dencorr__control_refraction_test_1088281088_minus_refraction_test_1088281088_2dkpower

I will note that the calibration solutions are exactly the same, so the changes must come further down the pipe. This is expected given how refraction is applied. 

Happy to say this test passed.

PR #320, opened 2024-03-13
dictionary updates

bhazelton:
I'd like to get these in, we need them! @dstorer and @PyxieLouStar maybe you can expand the description on `no_frequency_flagging`? It's a little terse at the moment.


nicholebarry (commented):
`fill_model_visibilities` is hardcoded to 1 in the model generation code. Do we want this to be changed on input?

Default for `beam_nfreq_avg` is 1 within the codebase. Default for `beam_nfreq_avg` is 16 within the `eor_wrapper_defaults`, which is perhaps were the confusion came from here.

nicholebarry (approved):

PR #325, opened 2025-01-17
Calibration misc

nicholebarry:
These are miscellaneous calibration changes:

1) Store the used bandpass in the cal structure. This is important if you want to reconstruct each step of calibration after a run has finished. Purely for metadata/storing/tracking purposes.
2) Remove NANs from the yrange calculation of calibration plots for better viewing. This is something that rarely happens, but if you are playing around with some simulations and doing some hand-flagging, it's possible.
3) Model transfer framework that can handle WODEN inputs. Some time/baseline matching needs to occur because WODEN makes the full-model without flagging. This is kept very generalized to work with previous versions of WODEN that had a JD offset (noon versus midnight). 

All changes have been confirmed to work in PS.

### DIFF
--- a/fhd_core/source_modeling/vis_model_transfer.pro
+++ b/fhd_core/source_modeling/vis_model_transfer.pro
@@ -54,10 +54,15 @@ function vis_model_transfer,obs,params,model_transfer
     endfor
 
     ; Find which option was more successful in matching the times
+    ; Error if no matching times found or if matching times does not obviously cover all data times
+    if (total(matched_times_opt1 < 1) EQ 0) and (total(matched_times_opt2 < 1) EQ 0) then $
+      message, 'ERROR: No matching times found between transferred model and data.'
     if total(matched_times_opt1 < 1) GT total(matched_times_opt2 < 1) then matched_times = matched_times_opt1
     if total(matched_times_opt1 < 1) LT total(matched_times_opt2 < 1) then matched_times = matched_times_opt2
+    if total(matched_times < 1) NE data_n_time then $
+      message, 'ERROR: Number of matching times found between transferred model and data does not equal the number of timesteps in data.'
 
-    endif else matched_times = indgen(model_n_time)
+  endif else matched_times = indgen(model_n_time)
 
   ;An arbitrary larger number than tiles to make an index array
   baseline_mod = ULONG(obs.n_tile*2)
@@ -73,6 +78,7 @@ function vis_model_transfer,obs,params,model_transfer
   endfor
 
   ; In each matched timestep, match the baselines and fill a new, matched model array 
+  ; Unmatched times in the transferred model are not included in the matched model array 
   for time_i=0, data_n_time-1 do begin
     ;suba is the subset of indices in the first array that are also in the second.
     ;subb is the subset of indices in the second array that are also in the first.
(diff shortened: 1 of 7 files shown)


fhd_core/source_modeling/vis_model_transfer.pro
```diff
@@ -11,6 +14,76 @@ function vis_model_transfer,obs,model_transfer
...
+    ; Option 2
+    matched_times_opt2 = intarr(data_n_time)
+    for data_time_i=0, data_n_time-1 do begin
+      min_val = min(abs(model_time - (data_Jdate[data_time_i]-time_res_Jdate)), min_ind)
+      if min_val LT tolerance then matched_times_opt2[data_time_i] = min_ind
+    endfor
+
+    ; Find which option was more successful in matching the times
+    if total(matched_times_opt1 < 1) GT total(matched_times_opt2 < 1) then matched_times = matched_times_opt1
+    if total(matched_times_opt1 < 1) LT total(matched_times_opt2 < 1) then matched_times = matched_times_opt2
```
bhazelton, line 58: what if neither are very good? should we error or warn if the best one is still way off?

fhd_core/source_modeling/vis_model_transfer.pro
```diff
@@ -11,6 +14,76 @@ function vis_model_transfer,obs,model_transfer
...
+
+  ; In each matched timestep, match the baselines and fill a new, matched model array 
+  for time_i=0, data_n_time-1 do begin
+    ;suba is the subset of indices in the first array that are also in the second.
+    ;subb is the subset of indices in the second array that are also in the first.
+    match, data_baseline_index[time_i*data_nbaselines:(time_i+1)*data_nbaselines-1], $
+      model_baseline_index[matched_times[time_i]*model_nbaselines:(matched_times[time_i]+1)*model_nbaselines-1], $
+      suba, subb
+    for pol_i=0, obs.n_pol-1 do (*matched_model[pol_i])[*,suba+time_i*(data_nbaselines)] = (*vis_model_arr[pol_i])[*,subb+matched_times[time_i]*model_nbaselines]
+  endfor
```
bhazelton, line 83: Ahh, so this takes care of the ones that don't match? what exactly happens if there are times that don't match?

bhazelton (commented):
This looks very sensible, I just have a couple questions about the time matching.

fhd_core/source_modeling/vis_model_transfer.pro
nicholebarry: I've included two errors to be explicit, one if no times matched and one if the times matched do not equal the number of timesteps in the data.

fhd_core/source_modeling/vis_model_transfer.pro
nicholebarry: I've added an extra comment to be a little more clear.

Any unmatched transferred model times are discarded. The idea here is that you have a full model for all times, and you are matching it to imperfect, flagged data. 

You may have noticed that I did not force timestep matching if the number of times is the same between the data and the model. I want to keep it somewhat flexible in case people are working with simulations. Ideally everyone fills in the metadata for simulations, but I'm not going to force them.

bhazelton (approved):
Thanks for the changes, this looks good to me now.

PR #326, opened 2025-01-17
remove float force for lat,lon,alt

### LINKED ISSUE #322: lat/lon/alt values are only saved as floats rather than doubles on the obs structure
This limits their accuracy at a level that is above the pyuvdata tolerances.

nicholebarry:
Fixes #322 

Pretty simple fix here, just take out the float casting when the lat/lon/alt are read in.

Small differences in the PS, but as expected. Values stored as doubles confirmed.


bhazelton (approved):
Yay! Thanks for this fix!!

Issue #311, opened 2023-08-25
`fhd_struct_init_meta` and `fhd_struct_init_obs` tile flag bug


SkyWa7ch3r:
While translating and testing for PyFHD I noticed the `tile_flag` variable inside `fhd_struct_init_meta` ends up with indexes beyond the array during it's creation furthermore, it ends up flagging tiles different for different polarizations. the end result is in `fhd_struct_init_obs`  when applying the `meta.tile_flag` to `tile_use` on 2 polarizations we end up with flagged tiles, while on 4 polarizations we end up with no flagged tiles. 

For now in PyFHD I apply the `single_i` to the `flag` column inside the `metafits` file and this gets read into the `meta['tile_flag']` dictionary value which then gets applied to `obs['baseline_info']['tile_use'] = 1 - meta['tile_flag']`, is this a suitable approach that could be done in FHD as well?

This issue can be tested with the observation `1088716176`, which is one of the observations used to test PyFHD via WODEN simulations or otherwise.

nicholebarry:
Fixed with #327 . Will need to add to PyFHD. No expected issues there, should be pretty simple

PR #327, opened 2025-01-17
convert tile flag to cross-corr flags

nicholebarry:
This one is bit complicated. This is in response to issue #311 .

The suggested fix in #311 could be more generalized. For example, the metadata file is actually structured such that the Y pol is before the X pol (why?!?!). Even though we currently flag XX, YY, XY, and YX the same *currently* doesn't mean we always will want that. Therefore, I've constructed a way that is generalized.

Essentially, the code grabs the individual dipole flags and matches their metadata name (i.e. X) to a component in the cross-correlation name (i.e. XY). Then a logical OR is performed to get the cross-correlation flag. 

So, an X-dipole flag will end up flagging XX, XY, and YX. But not YY.

No difference in the 2-pol. Only difference is in 4-pol flagging.

This will need to be implemented in pyFHD (low-priority) -- I will make an issue and link it.


bhazelton (approved):
This seems good to me.

Issue #321, opened 2024-04-23
auto_ratio_calibration ignores time-dependent flagging


PyxieLouStar:
Autocorrelations are averaged in time before being used by `cal_auto_ratio.pro` to compute auto ratios. This time averaging, performed in `vis_extract_autocorr.pro`, uses the `time_use` field from the `obs` structure to select which time stamps to include. However, the `time_use` field does not contain any information from time-dependent flagging present in the uvfits file. Thus time stamps which have been flagged in preprocessing (such as by SSINS, for example) are included in the average (and thus also in the final calibration solutions) when they should be excluded.

nicholebarry:
Fixed with #328 . Will need to add to PyFHD. No expected issues there

PR #328, opened 2025-01-19
Proper tracking of flagged times

nicholebarry:
This fix addresses issue #321, where the auto calibration was using times which should have been flagged.

This underlying root cause is that the time_use array was not checked against the weights array. In the case of SSINS, the weights for an entire time step will flagged, but this was not propagating to the time_use array in FHD. The time_use array is used a few circumstances, the two major instances being 1) auto calibration averaging and 2) noise calculation. Given that this affects the results in a few places, there is an overall change in the PS.

Below is a test against a control for an observation that had 4 SSINS flagged times. Differences are small and somewhat hard to interpret. Not sure if we want to pursue this differences deeper or if we are happy to take the fact that this is the proper time_use handling. 

fhd_nb_ntime_noimgclip_dft_averemove_swbh_dencorr__control2_1088281088_minus_test2_1088281088_2dkpower
fhd_nb_ntime_noimgclip_dft_averemove_swbh_dencorr__control2_1088281088_diffratio_test2_1088281088_2dkpower

### DIFF
--- a/fhd_core/visibility_manipulation/vis_flag_basic.pro
+++ b/fhd_core/visibility_manipulation/vis_flag_basic.pro
@@ -126,10 +126,20 @@ FOR ti=0L,n_time-1 DO BEGIN
         ENDFOR
     ENDIF
     ; Make sure the time_use array matches the flagging of the visiblities
-    FOR pol_i=0, n_pol-1 DO BEGIN
-        ; If any polarization is fully flagged, then count that time as flagged
-        time_use[ti] *= total((*vis_weight_ptr[pol_i])[*, bin_offset[ti]:bin_offset[ti+1]-1] > 0) GT 0
-    ENDFOR
+    IF STRMID(instrument,0,3) eq 'mwa' THEN BEGIN
+        FOR pol_i=0, n_pol-1 DO BEGIN
+            ; If any polarization is fully flagged, then count that time as flagged for the MWA
+            time_use[ti] *= total((*vis_weight_ptr[pol_i])[*, bin_offset[ti]:bin_offset[ti+1]-1] > 0) GT 0
+        ENDFOR
+    ENDIF ELSE BEGIN
+        ; For other instruments, do not assume that the flagging of the visibilities is the same for all polarizations
+        time_use_temp = 0
+        FOR pol_i=0, n_pol-1 DO BEGIN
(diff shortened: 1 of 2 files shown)


fhd_core/visibility_manipulation/vis_flag_basic.pro
```diff
@@ -112,22 +112,27 @@ IF Keyword_Set(no_frequency_flagging) THEN BEGIN
...
+    ; Make sure the flagging of the visibilities matches the time_use array  
+    IF time_use[ti] LE 0 THEN BEGIN 
+        FOR pol_i=0, n_pol-1 DO BEGIN  
+            (*vis_weight_ptr[pol_i])[*, bin_offset[ti]:bin_offset[ti+1]-1] = 0
+        ENDFOR
+    ENDIF
+    ; Make sure the time_use array matches the flagging of the visiblities
+    FOR pol_i=0, n_pol-1 DO BEGIN
+        ; If any polarization is fully flagged, then count that time as flagged
+        time_use[ti] *= total((*vis_weight_ptr[pol_i])[*, bin_offset[ti]:bin_offset[ti+1]-1] > 0) GT 0
```
bhazelton, line 131: This might need to be revisited in the case that Dara developed for HERA where antenna flagging can be per pol. I don't think we want to support different per-time flagging per pol if both pols are included, but we do want to support entirely flagging one pol of an antenna and not flagging the other pol. We can push that to another PR if we want just wanted to note the potential conflict.

bhazelton (commented):
Thank you for hunting this down!! I'm not worried about the PS changes, they look small and reasonable and this is clearly the more correct thing to do. I have one comment about the polarization aspect that doesn't need to block this but I want to understand a little better.

bhazelton:
@miguelfmorales you should see this.

nicholebarry:
More than happy to account for as many instruments as possible in development! I've added a simple fix for this instance, let me know if that satisfies the needs.

Essentially, 1) keep flags the same per pol per timestep for MWA, and 2) flag the timestep only if all pols are flagged for non-MWA.

bhazelton (approved):
Thanks for the changes, I think this looks good. I think we may need to revisit this when we merge in the other per-pol flagging stuff.

PR #329, opened 2025-01-23
Reducing skymodel confusion

nicholebarry:
Some skymodel behaviour was very opaque, as pointed out by Shijiao. So, I set out to make it a little more obvious what is going on.

**Previous functionality**
When `return_cal_visibilities` is set, then the calibration visibilities are kept around after calibration. If a different set of visibilities need to be calculated for the subtraction model, only sources that are *extra* to the initial calibration sources list are calculated and added to the returned calibration visibilities. This assumes that the calibration sources are a subset of the model sources. While this reduces the number of calculations, it is not obvious from the user that setting `return_cal_visiblities` would cause this behaviour.

The skymodel that is saved, regardless of whether or not `return_cal_visibilities` is set, only includes model sources that are not "duplicates" of the calibration sources. There is no indication in the source list whether a source was used in the model. Assumptions include 1) that the model is a subset, and 2) that sources that are co-located to within 1/100th of a pixel are duplicates, regardless of brightness.

**New functionality**
When `return_cal_visibilities` is set, then the calibration visibilities are kept around after calibration. If a different set of visibilities need to be calculated for the subtraction model, all will be calculated unless `combine_skymodels` is set. This gives flexibility to do the previous functionality (and be more efficient in your number of calculations) or do something more complicated where the assumptions do not apply.

When `combine_skymodels` is set, the code attempts to combine the skymodels as above. Extra model sources are marked with unsetting the `include_calibration` keyword (an already present keyword) for that source in the skymodel; duplicates are not included. If it seems as though the model is not a subset of calibration, then all sources are saved. 

If `combine_skymodels` is not set, then code will save separate skymodels.

`save_skymodels` must be set for skymodels to be saved.

### DIFF
--- a/dictionary.md
+++ b/dictionary.md
@@ -225,7 +225,6 @@ This is a work in progress; please add keywords as you find them in alphabetical
   -*Default*: equal to `max_baseline` <br />
 
 **max_calibration_sources**: limits the number of sources used in the calibration. Sources are weighted by apparent brightness before applying the cut. Note that extended sources with many associated source components count as only a single source. <br />
-  -*Dependency*: The sources are also included in the model if `return_cal_visibilities` is set. <br />
   -*Default*: All valid sources in the catalog are used. <br />
 
 **min_cal_baseline**: the minimum baseline length in wavelengths to be used in calibration. <br />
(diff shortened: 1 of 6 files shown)


bhazelton (commented):
This looks like a really good improvement, much more predictable behavior.  one question I have is how does this interact with the current "standard" run? I feel like it doesn't -- the standard run doesn't use a different set of sources for calibration vs subtraction. Mostly wondering who all needs to be informed of this change.

nicholebarry:
You are correct -- this won't affect a standard run. The only person I know of who will be affected at the moment is Shijiao (who alerted me to the issue) and I will let her know.

bhazelton (approved):

PR #330, opened 2025-01-24
Beam misc

nicholebarry:
A few little tweaks to the beam setup:

1) Better beam transfer handling. Actually attempts to match the group id* between the transferred beam and the observation via the unique baseline identifier. If it fails, then it sets the group id to the first one (standard style run)

*(aka the identifier which says which unique beam it is. If there is only one unique beam, then all baselines have the same group id)

2) Slight increase in the loop for gridding each baseline with Jack's uvw phasing in image space. Only helped by about 5%--10%, so not a huge improvement. 
 
3) Add dipole gain handling from the metafits. This is a simulation-based input. Could be expanded to make unique beams (future work).

4) The square beam volume for the gaussian decomposition of the beam. The square beam volume is part of the calculation for our observation volume, so extremely important. We calculate the analytic volume of a set of gaussians. This is actually an overestimate though due to the horizon edge and the beam-clip-and-renormalize option. While I've left the skeleton of the analytic volume in the code, I reverted the square beam volume calculation to the discrete estimate, which seems more accurate.

### DIFF
--- a/fhd_core/setup_metadata/fhd_struct_init_meta.pro
+++ b/fhd_core/setup_metadata/fhd_struct_init_meta.pro
@@ -1,4 +1,4 @@
-FUNCTION fhd_struct_init_meta,file_path_vis,hdr,params,layout,lon=lon,lat=lat,alt=alt,n_tile=n_tile,$
+FUNCTION fhd_struct_init_meta,file_path_vis,hdr,params,lon=lon,lat=lat,alt=alt,n_tile=n_tile,$
     zenra_in=zenra_in,zendec_in=zendec_in,obsra_in=obsra_in,obsdec_in=obsdec_in,phasera_in=phasera_in,phasedec_in=phasedec_in,$
     rephase_to_zenith=rephase_to_zenith,precess=precess,degpix=degpix,dimension=dimension,elements=elements,$
     obsx=obsx,obsy=obsy,instrument=instrument,mirror_X=mirror_X,mirror_Y=mirror_Y,no_rephase=no_rephase,$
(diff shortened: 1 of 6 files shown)


fhd_core/setup_metadata/fhd_struct_init_meta.pro
```diff
@@ -82,7 +95,7 @@ ENDIF ELSE BEGIN
     hist_A1=histogram(tile_A1,min=1,max=n_tile,/binsize,reverse_ind=ria)
     hist_B1=histogram(tile_B1,min=1,max=n_tile,/binsize,reverse_ind=rib)
     hist_AB=hist_A1+hist_B1
-    tile_names=layout.antenna_numbers
+    tile_names=indgen(n_tile)+1
```
bhazelton, line 98: This seems wrong to me, I thought that tile_names in the FHD metadata reflected the true tile names (e.g. starting at 11) not the index...

bhazelton (commented):
I'm not sure I follow all of this, but what I do follow mostly seems reasonable, although I have one question about the handling of tile_names.

fhd_core/setup_metadata/fhd_struct_init_meta.pro
nicholebarry: Good catch, that was an accidental push

nicholebarry:
Thanks for catching the tile names. That was a bad hotfix because the layout structure was not being passed in. Layout structure is now being passed in, and bad hotfix removed. 

I also found another bad hotfix regarding the group_id in gridding. Reverted. Tested and works.

bhazelton (dismissed):
This looks good to me as far as I can follow it.

bhazelton (approved):

Issue #336, opened 2026-02-01
Gain solver stops too early, results in bad calibration


baron-de-montblanc:
I've identified an issue in `fhd_core/calibration/vis_calibrate_subroutine.pro`. Specifically, several of my observations (MWA Fall 2016 season) are triggering the following breakpoint (lines 257-266):

```idl
; Halt if the strict convergence is worse than most of the recent iterations
divergence_test_1 = convergence_strict ge median(conv_test[fii, i - divergence_history - 1 : i - 1])
; Also halt if the convergence gets significantly worse in one iteration
divergence_test_2 = convergence_strict ge min(conv_test[fii, 0 : i - 1]) * divergence_factor
if divergence_test_1 or divergence_test_2 then begin
     ; If both measures of convergence are getting worse, we need to stop.
     print, string(format = '("Calibration diverged at iteration ", A, " for pol_i: ", A, " freq_i: ", A,". Convergence was: ", A, " threshold was: ", A)', $
                  Strn(i), Strn(pol_i), Strn(fi), Strn(conv_test[fii, i - 1], format = '(e12.2)'), Strn(conv_thresh, format = '(e12.2)'))
     divergence_flag = 1
     break
``` 

This is happening early on, at iteration ~10, for _all_ tiles, but only certain frequencies. Here is an example of the resulting `cal_amp.png`:

<img width="1024" alt="Image" src="https://github.com/user-attachments/assets/64aacf15-ccf7-4c91-9971-398b47ad41a9" />

I investigated the gain vs. iteration history for the same OBSID (1161091512):

<img width="1920" height="641" alt="Image" src="https://github.com/user-attachments/assets/b04409e3-5159-4f9a-be4d-713496ed79a7" />

Now if I comment out the `divergence_flag = 1` and `break` on lines 265-266, and run the same calibration job, we see that frequency 111 is actually able to converge:

<img width="1920" height="683" alt="Image" src="https://github.com/user-attachments/assets/b145d8fb-7e86-4a11-8c86-3602fa0630b9" />

`divergence_test_1` and `divergence_test_2` are entirely dependent on parameters hard-coded in lines 22-23:

```idl
divergence_history = 3 ; halt if the strict convergence is worse than most of the last x iterations
divergence_factor = 1.5 ; halt if the convergence gets significantly worse by a factor of x in one iteration
``` 

So maybe a solution would be to lower these thresholds. Happy to test different values and open to any other suggestions!

bhazelton:
[...]

PR #338, opened 2026-02-10
divergence keyword flexibility

nicholebarry:
In response to issue #336 .

Two keywords, `divergence_history` (halt after x iterations if things get worse by a factor of y) and `divergence_factor` (if the gains are worse than factor y after x iterations, then halt), are currently hardcoded in the calibration code. I've added flexibility to set them from your input keywords. 

In particular, `divergence_history` was set to 3. Divergence is checked after an initial set of iterations just for phase (usually defaults to 4 in standard MWA) and a full cycle of `divergence_history`. So, divergence is usually checked at iteration 8. Set this higher to delay divergence checks. 

`divergence_factor` is set to 1.5. This probably doesn't need to be changed for decent MWA obs. 

This will also be really useful to calibrating other instruments, like LOFAR.


baron-de-montblanc:
I tried this fix and while it does seem to prevent early stopping and allow the calibration to converge, I don't get the same calibration solutions as when I simply comment out `divergence_flag = 1` and `break` on lines 265-266 of `fhd_core/calibration/vis_calibrate_subroutine.pro`. My version (working off of the `stokesI_hack` branch as I typically do):
my_version
When using the new fix with `divergence_history = 15` (working off of this PR branch):
nicholes_fix
I'm thinking the discrepancy is coming from a combination of using different branches and the modification of the `divergence_history` keyword. Is it worth trying to hunt this down? If so, any suggestions on what I should be looking for?

PR #344, opened 2026-04-02
Preserve double precision in gridding

bhazelton:
I was trying to get pyFHD degridding to agree with FHD degridding and after a number of fixes they were agreeing to a part in 10^4, which didn't seem good enough. I ran the codes in parallel and compared various values until I found that the gridding kernels after interpolation were not agreeing to machine precision. Looking carefully at that code I found that the baseline centers (relative to the psf grid) were being explicitly cast to floats rather than being preserved as doubles. When I removed this cast, pyFHD's degridded visibilities agree with FHD to (double) machine precision.

I'm wondering why this cast is there, was there a concern about speed or memory use? I think we have repeatedly found that double precision is required, so I'd like to remove this casting unless there's a really compelling reason to leave them in.

### DIFF
--- a/fhd_core/gridding/baseline_grid_locations.pro
+++ b/fhd_core/gridding/baseline_grid_locations.pro
@@ -99,8 +99,8 @@ Function baseline_grid_locations,obs,psf,params,n_bin_use=n_bin_use,bin_i=bin_i,
   ENDIF
 
   ; Center of baselines for x and y in units of pixels
-  xcen=Float(frequency_array#Temporary(kx_arr))
-  ycen=Float(frequency_array#Temporary(ky_arr))
+  xcen=frequency_array#Temporary(kx_arr)
+  ycen=frequency_array#Temporary(ky_arr)
 
   ; Pixel number offet per baseline for each uv-box subset 
   x_offset=Fix(Floor((xcen-Floor(xcen))*psf_resolution) mod psf_resolution, type=12) ; type=12 is unsigned int


fhd_core/gridding/baseline_grid_locations.pro
```diff
@@ -99,8 +99,8 @@ Function baseline_grid_locations,obs,psf,params,n_bin_use=n_bin_use,bin_i=bin_i,
   ENDIF
 
   ; Center of baselines for x and y in units of pixels
-  xcen=Float(frequency_array#Temporary(kx_arr))
-  ycen=Float(frequency_array#Temporary(ky_arr))
+  xcen=frequency_array#Temporary(kx_arr)
+  ycen=frequency_array#Temporary(ky_arr)
```
isullivan, line 103:
```suggestion
  xcen=Double(frequency_array#Temporary(kx_arr))
  ycen=Double(frequency_array#Temporary(ky_arr))
```

isullivan (approved):
Some of the structure metadata are strongly typed, so you can get an error if these are sometimes `float` and sometimes `double` (or worse, `int`). We tried to make everything we thought could be a `float` forced to that type for speed and memory, but if this needs to be `double` then it should always be explicitly forced to `double`.

bhazelton:
> Some of the structure metadata are strongly typed, so you can get an error if these are sometimes `float` and sometimes `double` (or worse, `int`). We tried to make everything we thought could be a `float` forced to that type for speed and memory, but if this needs to be `double` then it should always be explicitly forced to `double`.

Ok. In the examples I've looked at, the arrays are doubles coming into this point. I'm not sure if they should be cast to doubles here or earlier, I'll try to look at it and figure that out.

bhazelton:
> Some of the structure metadata are strongly typed, so you can get an error if these are sometimes `float` and sometimes `double` (or worse, `int`). We tried to make everything we thought could be a `float` forced to that type for speed and memory, but if this needs to be `double` then it should always be explicitly forced to `double`.

None of the derivations of this calculation are stored in structures. They are used in calculating various quantities, but those quantities are always inserted into arrays with pre-defined types, so I don't think the typing of them has any affect on anything that is stored in structures. 

Given that, should I do an explicit cast to double or should I leave it to inherit the type of the input arrays? I'm inclined towards the latter unless we want to move towards always doing everything at double precision.

isullivan:
> Given that, should I do an explicit cast to double or should I leave it to inherit the type of the input arrays? I'm inclined towards the latter unless we want to move towards always doing everything at double precision.

It is probably fine to let it inherit the type of the input arrays. I've been trying to remember why that cast to `float` was there in the first place, and I think it was to prevent using double precision, for speed and memory considerations. So leaving it unspecified seems OK.

PR #346, opened 2026-04-09
Add support for double precision HEALPix cubes

isullivan:


bhazelton:
merged via #347 (which built off of this)

PR #347, opened 2026-04-13
HEALPix cubes can now be saved out in double precision

baron-de-montblanc:
This is an extension to PR #346, fixing a small bug in `healpix_snapshot_cube_generate.pro` where cubes were initialized using `fltarr`, disregarding the `double_precision` keyword. Cubes are now initialized as either `fltarr` or `dblarr` depending on the value of `double_precision`, enabling them to be saved out in double precision.


bhazelton (approved):
Looks good to me!

PR #348, opened 2026-07-27
transfer_eor hotfix

nicholebarry:
Better keyword naming. Follows other format like the model. 

I also kept on having problems with keyword confounding, which can happen if you have a keyword that is similar at the start.


Tiasha-26 (approved):
maybe 'transfer_eor_file' a better keyword

Nazeef-Noorie (approved):

PR #350, opened 2026-07-27
h5 option for data input

nicholebarry:
Can now read h5 files for input. This has yet to be extended to the model transfer, though hypothetically this is easily doable.

I am not super happy with the naming of things. The h5 reader is internal to the `uvfits_read` to have as little footprint as possible. Perhaps we should change the name of `uvfits_read` to be more general. 

Reading data is easy enough, but generating the metadata in the correct format required bespoke code. 

One thing of note. IDL will not read h5 files that have been compressed. It is generally default to compress h5 files in pyuvdata, so `data_compression=None,flags_compression=None,nsample_compression=None` must be set. I need to think of a good way to warn the user, but haven't found a good way.

### DIFF
--- a/fhd_core/setup_metadata/fhd_struct_init_obs.pro
+++ b/fhd_core/setup_metadata/fhd_struct_init_obs.pro
@@ -9,13 +9,11 @@ FUNCTION fhd_struct_init_obs,file_path_vis,hdr,params,layout,dimension=dimension
     grid_nfreq_avg=grid_nfreq_avg,_Extra=extra
 ;initializes the structure containing frequently needed parameters relating to the observation
 IF N_Elements(instrument) EQ 0 THEN instrument='mwa' ELSE instrument=StrLowCase(instrument)
-obsname=file_basename(file_basename(file_path_vis,'.uvfits',/fold_case),'_cal',/fold_case)
+obsname=file_basename(file_basename(file_path_vis,'.uvfits',/fold_case),'.h5',/fold_case)
 git,'describe',result=code_version,repo_path=rootdir('fhd'),args='--long --dirty'
 IF N_Elements(code_version) GT 0 THEN code_version=code_version[0] ELSE code_version=''
 
 IF N_Elements(n_pol) EQ 0 THEN n_pol=hdr.n_pol
-pol_names=['XX','YY','XY','YX','I','Q','U','V']
-
 IF Tag_exist(layout, "n_antenna") THEN BEGIN
     n_tile=layout.n_antenna 
 ENDIF ELSE BEGIN
@@ -162,7 +160,7 @@ params.antenna1 = tile_A
 params.antenna2 = tile_B
 
 meta=fhd_struct_init_meta(file_path_vis,hdr,params,layout,degpix=degpix,dimension=dimension,elements=elements,$
-    n_tile=n_tile,instrument=instrument,pol_names=pol_names,meta_data=meta_data,meta_hdr=meta_hdr,_Extra=extra)
+    n_tile=n_tile,instrument=instrument,meta_data=meta_data,meta_hdr=meta_hdr,_Extra=extra)
 
 IF N_Elements(meta_data) EQ 0 THEN meta_data=Ptr_new() ELSE meta_data=Ptr_new(meta_data)
 IF N_Elements(meta_hdr) EQ 0 THEN meta_hdr=Ptr_new() ELSE meta_hdr=Ptr_new(meta_hdr)
@@ -181,6 +179,7 @@ FOR ti=0,N_Elements(time_cut)<2-1 DO BEGIN
     IF ti_end GE ti_start THEN time_use[ti_start:ti_end]=0
 ENDFOR
 n_time_cut = n_time - Total(time_use)
+
 tile_use1=intarr(n_tile)
 FOR pol_i=0,n_pol-1 DO BEGIN
     tile_use_i=where(*(meta.tile_flag[pol_i]) EQ 0,n_use)
@@ -206,7 +205,7 @@ IF dimension GT 4096 THEN BEGIN
         double_precision=1
     ENDIF
 ENDIF
-
+pol_names=['XX','YY','XY','YX','I','Q','U','V']
 healpix={nside:Long(nside),ind_list:String(ind_list),n_pix:Long(n_hpx),n_zero:Long(n_zero_hpx)}
 
 arr={tile_A:Long(tile_A),tile_B:Long(tile_B),bin_offset:Long(bin_offset),Jdate:meta.Jdate,freq:Double(frequency_array),fbin_i:Long(freq_bin_i),$
@@ -216,7 +215,7 @@ struct={code_version:String(code_version),instrument:String(instrument),obsname:
     kpix:Float(kbinsize),degpix:Float(degpix),obsaz:meta.obsaz,obsalt:meta.obsalt,obsra:meta.obsra,obsdec:meta.obsdec,$
     zenra:meta.zenra,zendec:meta.zendec,obsx:meta.obsx,obsy:meta.obsy,zenx:meta.zenx,zeny:meta.zeny,$
     phasera:meta.phasera,phasedec:meta.phasedec,orig_phasera:meta.orig_phasera,orig_phasedec:meta.orig_phasedec,$
-    n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:Long(n_time_cut),$
+    n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:n_time_cut,$
     n_vis:Long(n_vis),n_vis_in:Long(n_vis_in),n_vis_raw:Long(n_vis_raw),nf_vis:Long(n_vis_arr),primary_beam_area:Ptrarr(4),primary_beam_sq_area:Ptrarr(4),pol_names:pol_names,$
     jd0:meta.jd0,max_baseline:Double(max_baseline),min_baseline:Double(min_baseline),delays:meta.delays,lon:meta.lon,lat:meta.lat,alt:meta.alt,$
     freq_center:Float(freq_center),freq_res:Float(freq_res),time_res:Float(meta.time_res),astr:meta.astr,alpha:Float(spectral_index),$
--- a/fhd_core/setup_metadata/uvh5_read_layout.pro
+++ b/fhd_core/setup_metadata/uvh5_read_layout.pro
@@ -1,0 +1,48 @@
+FUNCTION uvh5_read_dataset, fid, path
+    did = H5D_OPEN(fid, path)
+    data = H5D_READ(did)
+    H5D_CLOSE, did
+    RETURN, data
+END
+
+FUNCTION uvh5_read_layout, file_path_vis, _EXTRA=extra
+
+    fid = H5F_OPEN(file_path_vis)
+
+    ant_names     = uvh5_read_dataset(fid, '/Header/antenna_names')
+    ant_numbers   = uvh5_read_dataset(fid, '/Header/antenna_numbers')
+    ant_positions = uvh5_read_dataset(fid, '/Header/antenna_positions')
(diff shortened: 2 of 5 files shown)


fhd_core/visibility_manipulation/uvh5_mrdfits.pro
```diff
@@ -0,0 +1,258 @@
...
+
+  obsname = FILE_BASENAME(filename)
+  IF STRMATCH(STRLOWCASE(obsname), '*.uvh5') THEN $
+      obsname = FILE_BASENAME(obsname, '.uvh5')
+  IF STRMATCH(STRLOWCASE(obsname), '*.h5') THEN $
+      obsname = FILE_BASENAME(obsname, '.h5')
+
+  data_header0 = STRARR(256)
+
+  SXADDPAR,data_header0,'NAXIS',7
```
bhazelton, line 135: This approach of effectively building a uvfits header from the hdf5 file feels very clunky to me. Why not just build the header structure that FHD wants from the hdf5 file like you did with the layout?

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -216,7 +215,7 @@ struct={code_version:String(code_version),instrument:String(instrument),obsname:
     kpix:Float(kbinsize),degpix:Float(degpix),obsaz:meta.obsaz,obsalt:meta.obsalt,obsra:meta.obsra,obsdec:meta.obsdec,$
     zenra:meta.zenra,zendec:meta.zendec,obsx:meta.obsx,obsy:meta.obsy,zenx:meta.zenx,zeny:meta.zeny,$
     phasera:meta.phasera,phasedec:meta.phasedec,orig_phasera:meta.orig_phasera,orig_phasedec:meta.orig_phasedec,$
-    n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:Long(n_time_cut),$
+    n_pol:Fix(n_pol,type=2),n_tile:Long(n_tile),n_tile_flag:Long(n_flag),n_freq:Long(n_freq),n_freq_flag:0L,n_time:Long(n_time),n_time_flag:n_time_cut,$
```
bhazelton, line 218: why remove the Long here?

fhd_core/visibility_manipulation/uvfits_read.pro
```diff
@@ -24,16 +24,28 @@ ENDIF ELSE BEGIN
     ENDIF
     
     t_readfits=Systime(1)
-    lun = fxposit(file_path_vis, 0,/readonly)
-    data_struct=mrdfits(lun,0,data_header0,/silent)
+    ext = STRLOWCASE(FILE_BASENAME(file_path_vis))
+    if STRMATCH(ext, '*.uvh5') OR STRMATCH(ext, '*.h5') THEN BEGIN
```
bhazelton, line 28: Maybe do this strmatch once and record the value so if we decide to update this in the future we only have to change it one place?

fhd_core/visibility_manipulation/uvfits_read.pro
```diff
@@ -24,16 +24,28 @@ ENDIF ELSE BEGIN
...
+        if N_elements(freq_split) EQ 0 THEN BEGIN
+            data_struct = uvh5_mrdfits(file_path_vis, data_header0)
```
bhazelton, line 30: the name "uvh5_mrdfits" breaks my brain. Can we just call it "read_uvh5" or something similar?

fhd_core/setup_metadata/fhd_struct_init_obs.pro
```diff
@@ -9,13 +9,11 @@ FUNCTION fhd_struct_init_obs,file_path_vis,hdr,params,layout,dimension=dimension
     grid_nfreq_avg=grid_nfreq_avg,_Extra=extra
 ;initializes the structure containing frequently needed parameters relating to the observation
 IF N_Elements(instrument) EQ 0 THEN instrument='mwa' ELSE instrument=StrLowCase(instrument)
-obsname=file_basename(file_basename(file_path_vis,'.uvfits',/fold_case),'_cal',/fold_case)
+obsname=file_basename(file_basename(file_path_vis,'.uvfits',/fold_case),'.h5',/fold_case)
```
bhazelton, line 12: This call confuses me, it's not obvious what it's doing with both a ".uvfits" and a ".h5" in the same line. It's probably something clever but I find it hard to read. Also, why only ".h5" and not ".uvh5"? And is the "_cal" that got removed never useful?

fhd_core/setup_metadata/uvh5_read_layout.pro
```diff
@@ -0,0 +1,48 @@
...
+    nant = N_ELEMENTS(ant_numbers)
+
+    ; IDL/HDF5 dimension order can be reversed, so make sure positions are [3, nant].
+    pos_dims = SIZE(ant_positions, /DIMENSIONS)
+    IF pos_dims[0] NE 3 THEN ant_positions = TRANSPOSE(ant_positions)
+
+    ; Build a layout structure with the fields FHD usually needs.
+    ; UVH5 antenna_positions are ECEF positions relative to the telescope location,
+    ; not an AIPS AN binary table. pyuvdata documents that telescope.antenna_positions
+    ; are ECEF positions relative to telescope.location. :contentReference[oaicite:2]{index=2}
```
bhazelton, line 33: Yes, the antenna positions stored in uvh5 files are not in the same coordinate system as antenna positions stored in uvfits files. I think we need to handle that here so they're always in the same coordinate system inside FHD. I'm not sure how they're used currently within FHD, but when pyuvdata reads them back in from FHD it assumes they are in the uvfits coordinate system.

bhazelton (commented):
Thank you! Getting uvh5 files into FHD will make things so much easier. I totally agree about naming, let's rename it "uvfits_read" to be more general.

fhd_core/visibility_manipulation/uvh5_mrdfits.pro
nicholebarry: done!

fhd_core/setup_metadata/fhd_struct_init_obs.pro
nicholebarry: oops, put back

fhd_core/visibility_manipulation/uvfits_read.pro
nicholebarry: changed it so there was only one place where it needed to be checked.

fhd_core/visibility_manipulation/uvfits_read.pro
nicholebarry: done!

fhd_core/setup_metadata/fhd_struct_init_obs.pro
nicholebarry: ya, this is some gross IDL right here. Unfortunately, I think it's here to stay. Essentially the logic is "if uvfits is the extension, strip it, and then redo the logic for h5 or cal."

I don't even remember .cal files, so I figured that was an old option I could remove. 

Added functionality for uvh5, not just h5.

fhd_core/setup_metadata/uvh5_read_layout.pro
nicholebarry: Haven't implemented anything along those lines, partly because I don't know how to implement that. Happy to add an issue unless you have thoughts.

nicholebarry:
Functionality for partial reading of the h5 files has been added, where you can partial read in freq, time, pol, or antennas. This is to prepare for potential 256-tile data sets.

When partial reading is done, the output filename is modified to reflect that. This way, you can run multiple frequency runs at once and have separate files for each that can be combined later.

fhd_core/visibility_manipulation/read_uvh5.pro
```diff
@@ -0,0 +1,277 @@
...
+      if tag_exist(partial_read, 'antenna') then antenna_selection = LONG(partial_read.antenna)
+  ENDIF
+
+  IF N_ELEMENTS(freq_selection) EQ 0 THEN BEGIN
+      freq_indices = LINDGEN(nfreq_total)
+      freq_name_suffix = ''
+  ENDIF ELSE BEGIN
+      freq_indices = freq_selection
+      valid = WHERE((freq_indices GE 0) AND (freq_indices LT nfreq_total), nvalid)
+      IF nvalid EQ 0 THEN freq_indices = LONARR(0) ELSE freq_indices = freq_indices[valid]
```
bhazelton: Shouldn't there be an error here if you have no valid frequencies?  Same question for the other axes below...

fhd_core/visibility_manipulation/read_uvh5.pro
```diff
@@ -0,0 +1,277 @@
...
+
+      nsel = N_ELEMENTS(freq_selection)
+      IF nsel EQ 1 THEN BEGIN
+          freq_name_suffix = '_freq' + STRTRIM(freq_selection[0], 2)
+      ENDIF ELSE IF MAX(ABS(freq_selection[1:*] - freq_selection[0:nsel-2])) EQ 1 THEN BEGIN
+          freq_name_suffix = '_freq' + $
+              STRTRIM(MIN(freq_selection), 2) + '-' + STRTRIM(MAX(freq_selection), 2)
+      ENDIF ELSE BEGIN
+          freq_name_suffix = '_freq' + STRJOIN(STRTRIM(freq_selection, 2), '_')
+      ENDELSE
```
bhazelton: Shouldn't all this logic use freq_indices rather than freq_selection since that's what you're actually using to do the selection? Same question for the other axes below...

fhd_core/visibility_manipulation/read_uvh5.pro
```diff
@@ -0,0 +1,277 @@
...
+              STRTRIM(MIN(time_selection), 2) + '-' + STRTRIM(MAX(time_selection), 2)
+      ENDIF ELSE BEGIN
+          time_name_suffix = '_time' + STRJOIN(STRTRIM(time_selection, 2), '_')
+      ENDELSE
+  ENDELSE
+
+  IF N_ELEMENTS(antenna_selection) GT 0 THEN BEGIN
+      selected_times = time_indices[WHERE($
+          (ant1[time_indices] EQ antenna_selection[0]) OR $
+          (ant2[time_indices] EQ antenna_selection[0]), nselected_times)]
```
bhazelton: This looks like it's only going to pick out baseline-time indices if they involve the first antenna in `antenna_selection`. Should there be a loop over the antennas you want or some kind of array_match call to match all the antennas you want?

fhd_core/visibility_manipulation/read_uvh5.pro
bhazelton: Also, if this ends up dropping some things from the selection maybe you should warn the user?

fhd_core/visibility_manipulation/uvfits_read.pro
```diff
@@ -1,7 +1,8 @@
-PRO uvfits_read,hdr,params,layout,vis_arr,vis_weights,file_path_vis=file_path_vis,n_pol=n_pol,silent=silent,$
-    restore_vis_savefile=restore_vis_savefile,reorder_visibilities=reorder_visibilities,$
-    vis_time_average=vis_time_average,vis_freq_average=vis_freq_average,error=error,$
-    uvfits_spectral_dimension=uvfits_spectral_dimension,_Extra=extra
+PRO uvfits_read,hdr,params,layout,vis_arr,vis_weights,file_path_vis=file_path_vis,file_path_fhd=file_path_fhd,$
```
bhazelton: I'd slightly prefer something like "vis_read" or something else that's generalizes uvfits & uvh5

bhazelton (commented):
I added the logic to get antenna positions into the same frame as comes from uvfits.

I have a few questions about the data selection code.

I think there are ways to do partial reads with uvfits files as well, not sure if it's worth figuring it out at the moment.

bhazelton:
It'd be good to do a round trip test with this and pyuvdata to make sure all the parsing is correct. I think that'd look like:
FHD read uvh5
FHD write out unmodified (i.e. uncalibrated) visibilities
pyuvdata read in the uvh5 & the FHD written visibilities and assert they are the same.

fhd_core/setup_metadata/uvh5_read_layout.pro
bhazelton: I've added this handling.

PR #354, opened 2026-09-10
zoom window constraints

nicholebarry:
Component grouping was failing in some edge cases in simulation because there were no constaints on the zoom window. This pull request puts up guard rails to make sure edge cases are handled.


bhazelton (approved):
Looks reasonable to me

